PR #3, opened 2024-04-19
Initial user docs structure

imiller0:
Adding skeleton for initial user docs. Separation of user guide from reference configuration guide helps focus on different types of users.

Additional sections or guides can be added as needed.


pixelsoccupied:
/lgtm
/approve

Would be nice to add a job for link markdown e.g https://github.com/openshift/release/blob/master/ci-operator/config/openshift-kni/lifecycle-agent/openshift-kni-lifecycle-agent-main.yaml#L79-L85

PR #10, opened 2024-04-30
Converted to Separate Execution of Diff Command for Each CR

AlinaSecret:
This commit changes the run behavior of the diff command to execute separately for each CR instead of running the diff command once for all CRs. This change enables easy tracking of the number of CRs containing differences, and they are now added to the command output. Additionally, this change generates separate output for each CR, enabling the addition of additional information to each presented diff. For each found diff, the CR name and its correlated matching template name are now presented. Furthermore, a separator string has been added between each diff to allow clearer separation between CRs diff outputs. Also, the titles of the command summary output have been changed to be more informative.

### DIFF
--- a/pkg/compare/compare.go
+++ b/pkg/compare/compare.go
@@ -93,6 +93,8 @@ const (
 	noReffDirectoryWasPassed = "\"Reference directory is required\""
 	reffDirNotExistsError    = "\"Reference directory doesnt exist\""
 	emptyTypes               = "Templates don't contain any types (kind) of resources that are supported by the cluster"
+	DiffSeparator            = "**********************************"
+	ShowManagedFields        = false
 )
 
 type Options struct {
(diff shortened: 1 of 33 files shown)


pkg/compare/compare.go
```diff
@@ -357,19 +359,34 @@ func findAllRequestedSupportedTypes(supportedTypesWithGroups map[string][]string
...
 	differ, err := diff.NewDiffer("MERGED", "LIVE")
+	diffOutput := new(bytes.Buffer)
 	if err != nil {
-		return err
+		return diffOutput, err
 	}
 	defer differ.TearDown()
 
-	printer := diff.Printer{}
+	err = differ.Diff(obj, diff.Printer{}, ShowManagedFields)
```
ardaguclu, line 370: Perhaps, it would be useful to provide `ShowManagedFields` as a flag for the user, in case user wants to configure.

ardaguclu (commented):
Code looks good to me. Just a minor comment.

pkg/compare/compare.go
AlinaSecret: added the flag :)

PR #12, opened 2024-05-15
Add option to format output as JSON/YAML 

AlinaSecret:
A new flag '-o' ('output') is introduced, allowing the user to choose the output format of the command. The options are json/yaml/human. The default behavior remains the same, with output in 'human' format. The flag supports auto-completion for the output format options, implemented in the code of the command. To enable auto-completion when using the command as a plugin, an additional file 'kubectl_complete-cluster_compare' was added. The file should be present in a location in PATH along with the plugin to enable auto-completion. Also added supporting tests for JSON and YAML output modes.


AlinaSecret:
/hold

 depends on https://github.com/openshift/kube-compare/pull/10

pkg/compare/testdata/JSONOutput/localout.golden
```diff
@@ -0,0 +1 @@
+{"Summary":{"RequiredCRS":{"ExamplePart":{"Dashboard":["deploymentDashboard.yaml"]}},"NumMissing":1,"UnmatchedCRS":[],"NumDiffCRs":1},"Diffs":[{"DiffOutput":"diff -u -N TEMP/apps-v1_deployment_kubernetes-dashboard_dashboard-metrics-scraper TEMP/apps-v1_deployment_kubernetes-dashboard_dashboard-metrics-scraper\n--- TEMP/apps-v1_deployment_kubernetes-dashboard_dashboard-metrics-scraper\tDATE\n+++ TEMP/apps-v1_deployment_kubernetes-dashboard_dashboard-metrics-scraper\tDATE\n@@ -10,7 +10,7 @@\n   revisionHistoryLimit: 10\n   selector:\n     matchLabels:\n-      k8s-app: dashboard-metrics-scraper\n+      k8s-app: dashboard-metrics-scraper-diff\n   template:\n     metadata:\n       labels:\n","CorrelatedTemplate":"deploymentMetrics.yaml","CRName":"apps/v1_Deployment_kubernetes-dashboard_dashboard-metrics-scraper"}]}
```
pixelsoccupied, line 1: missing new line for some the files otherwise lgtm!

pkg/compare/compare.go
```diff
@@ -96,12 +99,21 @@ const (
 	DiffSeparator            = "**********************************"
 )
 
+const (
+	Json  string = "json"
+	Yaml         = "yaml"
+	Human        = "human"
```
pixelsoccupied, line 105: in this case "human" would be default output (what we currently have)?

AlinaSecret: yes

pixelsoccupied: Gotcha! So in that case can we treat this as "default" and just drop reference to "human"? It sort of reads like we are creating a new file type `.human` (like .json + .yaml)?

AlinaSecret: not sure i understood exactly your suggestion , but i added a default option that is equal to the human one

PR #13, opened 2024-05-16
Add release endpoint to Makefile

AlinaSecret:
This commit adds the `release` and `release-dry-run` endpoints to the Makefile. These endpoints use Goreleaser to build the compare binary for different architectures and platforms. The `release-dry-run` endpoint only builds the binaries, creates the archives, and generates the release notes. The `release` endpoint also uploads the release to GitHub. The release notes contain a list of all the changes from the previous tag to the newest tag.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,6 +2,10 @@
 ENGINE ?= docker
 IMAGE_NAME=kube-compare
 
+PACKAGE_NAME          := github.com/openshift/kube-compare
+GOLANG_CROSS_VERSION  ?= v1.22.3
+
+
 .PHONY: build
 build:
 	go build ./cmd/kubectl-cluster_compare.go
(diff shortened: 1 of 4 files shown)


AlinaSecret:
/cc @pixelsoccupied
/cc @imiller0 

here is an example to how it looks from my fork: https://github.com/AlinaSecret/kube-compare/releases/tag/v1.0.0
(because of the mergebot we have the merge commits in the change log, its very fast to remove them manually from the release notes so for now i think there is no need to spend time on doing it automatically)

Makefile
```diff
@@ -28,3 +32,27 @@ markdownlint: markdownlint-image  ## run the markdown linter
...
+.PHONY: release
+release:
+	@docker run \
+		--rm \
+		-e GITHUB_TOKEN=$(GITHUB_TOKEN) \
+		-e CGO_ENABLED=1 \
+		-v /var/run/docker.sock:/var/run/docker.sock \
+		-v `pwd`:/go/src/$(PACKAGE_NAME) \
+		-w /go/src/$(PACKAGE_NAME) \
+		ghcr.io/goreleaser/goreleaser-cross:${GOLANG_CROSS_VERSION} \
```
pixelsoccupied, line 57: Good stuff! Do mind adding a few lines on how to do the release? E.g what can i steps i have to take release v1.0.1? 

We could probably add `RELEASING.md` file to document it

AlinaSecret: added a release guide.
in general we need to discuss a little more about the versioning (not sure what is the tag for our first release)
but any way i dose not effect the PR .

Still haven't add the krew part, will add it once we do the first krew release, and then i will add the info on how to update it

PR #14, opened 2024-05-23
Add report-creator CLI tool

AlinaSecret:
Introduce report-creator, a CLI tool that allows generating a JUnit test
report from the JSON output of the 'kubectl cluster-compare' plugin. 

The tool categorizes the cluster comparison results into 3 test suites:

1. Diff Test Suite: Each test in this suite represents a matched CR compared
to a reference CR. 

2. Missing CRs Test Suite: Tests in this suite represent CRs absent in the
cluster but present in the reference.

3. Unmatched CRs Test Suite: This suite includes all CRs in the cluster not
matched to any reference.

Additionally, tests of the new tool are added to ensure the the report creation 
is as expected.  These tests use the same references used for the compare
 command tests to validate the alignment of the reporter with the compare 
command's reference and output format.


AlinaSecret:
/hold

depends on #12

PR #18, opened 2024-05-28
Change command behavior to skip invalid resources in local mode

AlinaSecret:
previously, when using the command in local mode with recursive mode enabled, the command would attempt to parse additional JSON/YAML files as resources. If any files were invalid, an error was presented to the user, and the compare results for other valid resources were not returned.

With this update, the command will now issue a warning for invalid resources instead of stopping execution. It will skip these problematic files and proceed to present the compare results for the valid resources.

### DIFF
--- a/pkg/compare/compare.go
+++ b/pkg/compare/compare.go
@@ -99,6 +99,9 @@ const (
 	refDirNotExistsError    = "\"Reference directory doesn't exist\""
 	emptyTypes              = "templates don't contain any types (kind) of resources that are supported by the cluster"
 	DiffSeparator           = "**********************************\n"
+	skipInvalidResources    = "Skipping %s Input contains additional files from supported file extensions" +
+		" (json/yaml) that do not contain a valid resource, error: %s.\n In case this file is " +
+		"expected to be a valid resource modify it accordingly. "
 )
 
 const (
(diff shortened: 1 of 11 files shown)


pkg/compare/compare.go
```diff
@@ -433,6 +443,14 @@ func (o *Options) Run() error {
 		return fmt.Errorf("failed to collect resources: %w", err)
 	}
 	r.IgnoreErrors(func(err error) bool {
+		if strings.Contains(err.Error(), "Object 'Kind' is missing") {
+			klog.Warningf(skipInvalidResources, extractPath(err.Error(), 3), "'Kind' is missing")
+			return true
+		}
```
pixelsoccupied, line 449: Instead of checking the error string we should check for the error type with As or better just use the helper function from apimachinery? https://github.com/kubernetes/apimachinery/blob/master/pkg/runtime/error.go#L93-L101

Same for the other check below

pkg/compare/testdata/InvalidResourcesAreSkipped/localerr.golden
```diff
@@ -0,0 +1,2 @@
+ 
+error code:1
```
pixelsoccupied, line 2: Missing an empty line

pkg/compare/compare.go
AlinaSecret: I wish i could,
the function from apimachinery doesn't work and also errors,As does not work, the type of the return error is:
errors.errorString so cant check for the error type
and the string is:
unable to decode "testdata/InvalidResourcesAreSkipped/resources/d1.json": Object 'Kind' is missing in '{
  "glossary": {
    "title": "example glossary"
      }
}'

pkg/compare/testdata/InvalidResourcesAreSkipped/localerr.golden
AlinaSecret: fixed

pkg/compare/compare.go
pixelsoccupied: Ah gotcha!

AlinaSecret:
@pixelsoccupied only fixed the tests from the lgtm (removed whitespace that was changed in previous merge)

PR #27, opened 2024-06-07
clarify the success messages during live mode

pixelsoccupied:
Success messages when running against live cluster should be clarified a bit more. 

Feel free to suggest any alt. Basically the goal is to be able to tell "All good" at a quick glance 

```
#prev

Summary
CRs with diffs: 1
No CRs are missing from the cluster
No CRs are unmatched to reference CRs
```

```
# cur 

Summary
Diff count: 1
Successfully found all required resources from the cluster
Successfully compared all reference resources
```

Additional update the gitignore to ignore local build

/cc @AlinaSecret  @nocturnalastro

### DIFF
--- a/pkg/compare/compare.go
+++ b/pkg/compare/compare.go
@@ -555,18 +555,18 @@ func newSummary(reference *Reference, c *MetricsCorelatorDecorator, numDiffCRs i
 func (s Summary) String() string {
 	t := `
 Summary
-CRs with diffs: {{ .NumDiffCRs }}
+Diff count: {{ .NumDiffCRs }}
 {{- if ne (len  .RequiredCRS) 0 }}
 CRs in reference missing from the cluster: {{.NumMissing}} 
 {{ toYaml .RequiredCRS}}
 {{- else}}
-No CRs are missing from the cluster
+Successfully found all required reference resources from the cluster
 {{- end }}
 {{- if ne (len  .UnmatchedCRS) 0 }}
 Cluster CRs unmatched to reference CRs: {{len  .UnmatchedCRS}}
 {{ toYaml .UnmatchedCRS}}
 {{- else}}
-No CRs are unmatched to reference CRs
+Successfully compared all reference resources
 {{- end }}
 `
 	var buf bytes.Buffer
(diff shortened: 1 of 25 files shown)


pkg/compare/compare.go
```diff
@@ -555,18 +555,18 @@ func newSummary(reference *Reference, c *MetricsCorelatorDecorator, numDiffCRs i
...
 {{- else}}
-No CRs are missing from the cluster
+Successfully found all required reference resources from the cluster
 {{- end }}
 {{- if ne (len  .UnmatchedCRS) 0 }}
 Cluster CRs unmatched to reference CRs: {{len  .UnmatchedCRS}}
 {{ toYaml .UnmatchedCRS}}
 {{- else}}
-No CRs are unmatched to reference CRs
+Successfully compared all reference resources
```
AlinaSecret, line 569: The message can be misleading, If there are entries in this section it means that there were no CRs in the cluster that don't have a match in the reference CRs
i think  if we want to phrase it in a positive way it should be : "Successfully found a match for all cluster CRs to reference CRs" 
(something like this , the phrasing\grammar can be changed but that is the idea) 

Anyway i think we should remove the section and include it only when using -A flag. when not using this flag this section will not be empty only if there is a CR that had multiple matches to refernce CRs, most of the time it will be empty and not clear enough. So its confusing, i will open some issues to address this behavior .

pkg/compare/compare.go
```diff
@@ -555,18 +555,18 @@ func newSummary(reference *Reference, c *MetricsCorelatorDecorator, numDiffCRs i
 func (s Summary) String() string {
 	t := `
 Summary
-CRs with diffs: {{ .NumDiffCRs }}
+Diff count: {{ .NumDiffCRs }}
```
AlinaSecret, line 558: I am a little scared that this may get confused with the number of lines with diffs, or number of diffs where each file can contain a number of diffs. its important that its clear that this i the number of cluster CRS that have diffs when compare to reference crs, (the important part is that its clear that its counting the number of CRS).
so i think we should make this part clear.

pixelsoccupied:
/close
Closing this for now. Will revisit this later after some discussion if needed!

PR #34, opened 2024-06-12
Remove jq from golangci-lint script and add more linters

nocturnalastro:


AlinaSecret:
Great! thanks  @nocturnalastro! always wanted to add linters but never found the time :)

PR #37, opened 2024-06-14
Merge reference info live before diff to get more relevent diffs

nocturnalastro:
# Background 
Jira: https://issues.redhat.com/browse/CNF-12172

~dependent on https://github.com/openshift/kube-compare/pull/38~


pixelsoccupied:
Could you link the JIRA/issues and update the PR description to explain what was done just so we understand the work in the future? 

Lgtm for me otherwise....I have created the new set DU profile rules based on this build https://github.com/openshift-kni/cnf-features-deploy/pull/1940

pixelsoccupied:
Might be good time to include PR template https://raw.githubusercontent.com/openshift-kni/lifecycle-agent/main/.github/pull_request_template.md

.golangci.yml
```diff
@@ -3,7 +3,7 @@ new: true
 new-from-rev: 722af08
 run:
   concurrency: 6
-  deadline: 5m
+  timeout: 5m
```
pixelsoccupied, line 6: This should be different PR?

nocturnalastro: Yeah I've split it off https://github.com/openshift/kube-compare/pull/38

pixelsoccupied:
Reached out to @imiller0 to talk about this change in behaviour and I think we can move forward as it drastically simplifies building the rules. But two things
- Allow bypassing of this behaviours with flag e.g `-v` `--show-more` ? 
- Document how to wrap a rule around to catch exceptions i.e catch/alert something that should be there. (I'm putting up an example of that in linked DU profile PR with PerformanceProfile)

AlinaSecret:
I think its better to include this behavior in the metadata.yaml, instead of as a flag, because it an attribute of the reference  the structure of the reference depends on this value.
and because its in the metadata.yaml we can also enable the feature only for a part/component in the reference,
what do you think? is it better to have a global setting or the option to configure it by  part?

AlinaSecret:
Also its important to add in this pr  in the docs an explanation about the merging behavior  
@nocturnalastro

AlinaSecret:
@pixelsoccupied 
not sure if i just didnt notice, but can you add the example you mentioned ( example of that in linked DU profile PR with PerformanceProfile)

pkg/compare/compare.go
```diff
@@ -450,9 +495,14 @@ func (o *Options) Run() error {
 			return err
 		}
 
+		localRef, patch, err := mergeAndGetPatch(localRef, clusterCR, o.showMore)
```
AlinaSecret, line 498: There is a dedicated section in the code for the merging part : https://github.com/openshift/kube-compare/blob/main/pkg/compare/compare.go#L499-L503
Its better to move it there, its pare of the diff interface  

just as an example  you can look a[t he implementation of the function in kubectl diff,](https://github.com/kubernetes/kubectl/blob/master/pkg/cmd/diff/diff.go#L346) they do a different merge but it shows the general idea

pkg/compare/compare.go
```diff
@@ -523,6 +574,7 @@ type DiffSum struct {
 	DiffOutput         string `json:"DiffOutput"`
 	CorrelatedTemplate string `json:"CorrelatedTemplate"`
 	CRName             string `json:"CRName"`
+	Patch              string
```
AlinaSecret, line 577: please remove patch from this PR, here in also in name of variables and other functions :)

pkg/compare/compare_test.go
```diff
@@ -149,6 +149,7 @@ type Test struct {
 	shouldDiffAll         bool
 	outputFormat          string
 	checks                Checks
+	showMore              bool
```
AlinaSecret, line 152: instead of a flag lets move it to be a field in the metadata.yaml
because the reference is built in a different way depending on the flag
also least include the option to disable/enable this option for separate CRs,  we can also do it for Components but as it looks to me in the example we have sometimes only on CR in  a component is more sensitive needing for an exact match.


I have another proposal to consider.:

adding for each template in the reference an option to specify for which sections to do the merge, here is an example:

```
Parts:
  - name: ExamplePart
    Components:
      - name: DemonSets
        type: Required
        requiredTemplates:
          - apps.v1.DaemonSet.kube-system.kindnet.yaml:
            MergeAttributes:
                   - - metadata
                     - annotations
                    - - metadata
                     - lables            
        optionalTemplates:
          - apps.v1.DaemonSet.kube-system.kindnet2.yaml

``` 
In case we want to use it for the whole file we can use "." or something else just wanted to open this derection) 
 I think this i very handy, that way CRs with problematic sections still can use the merges for some fields.
 This is the general idea. 
 
 summary: I think it should be in the metadata.yaml instead, and include an option for more options of configuration of the feature on different CRs   and we should do another pass on the reference for are case studies again to understand the situation where we would like an exact match

imiller0:
I think we need a bit more discussion on the merging behavior. There are some other cases where always merging will hide relevant differences that should be reviewed by the user for correctness. 
In the current implementation the reference designer can list specific fields in the metadata.yaml for suppressing differences (which is what the merge is also doing). I like the idea that the designer is explicitly saying "these fields are OK to differ" rather than defaulting to "ignore all differences unless explicitly written into the reference". Would augmenting the metadata.yaml support for suppressing differences help with this? Perhaps the ability to flag a CR as "ok to merge this whole CR", while others would continue to show more differences (because they aren't being merged).
The intent is that this tool is used to highlight risk due to differences between the user's config and the reference. The suppressing of differences is to help them focus on the important pieces by eliminating things we know are ok. If we aren't sure, or are going to make an error on what to show, we should error on the side of showing more and letting them review.

nocturnalastro:
@imiller0 

I've been working on some POCs for how to control this behaviour. Its very much doable, on a per template basis. I like the idea of having some config in the template via comments which allow you control if it is merged or not. [Here's an example of what I have in mind](https://github.com/nocturnalastro/kube-compare/pull/1/files#diff-2b57e592ec2cb72187c8b817d5888aafc8ad9be6985194710276ee5946f436d8R1)

The issue (as I see it) with listing explicit fields to omit is that it becomes very costly in enumerating them and will have substantial drift over time as the controllers change, making references a nightmare to maintain. Also `fieldsToOmit` would have to per-template for them to be useful for this. 

We could switch the behaviour to off by default instead but you will end up getting false positives with labels, annotations making it annoying to build references. Which is what @pixelsoccupied was facing when he was to building the DU profile.

While being able to save diffs and passing them in as acceptable goes some what to mitigate this I think would be annoying.

In some ways it can also make templates more explict if you really want a field never exist you can set it to null - showing you care about it not being there.

pkg/compare/testdata/ShowMoreFlag/reference/namespace.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: v1
```
nocturnalastro, line 1: TODO: make this have the merge-flag

pkg/compare/testdata/CheckProbesOnlyMatch/localout.golden
```diff
@@ -0,0 +1,38 @@
...
+-  command:
+-  - cat
+-  - /tmp/healthy
+   containers:
+   - args:
+     - /bin/sh
+     - -c
+     - touch /tmp/healthy; sleep 30; rm -f /tmp/healthy; sleep 600
+     image: registry.k8s.io/busybox
++    livenessProbe:
```
nocturnalastro, line 22: This template doesn't seem to be rendering correctly.

nocturnalastro (commented):
/hold 

Some of the tests aren't quite coming out corretly.

imiller0:
> @imiller0
> 
> I've been working on some POCs for how to control this behaviour. Its very much doable, on a per template basis. I like the idea of having some config in the template via comments which allow you control if it is merged or not. [Here's an example of what I have in mind](https://github.com/nocturnalastro/kube-compare/pull/1/files#diff-b425b3abdff419f70eb878f05febdc955e26605e8c823d76a645ee07a5b610b1R1)
> 
> The issue (as I see it) with listing explicit fields to omit is that it becomes very costly in enumerating them and will have substantial drift over time as the controllers change, making references a nightmare to maintain. Also `fieldsToOmit` would have to per-template for them to be useful for this.
Agreed that it would need to be per template (reference CR) to be accurate. One thing I like about your example is that the knowledge of how the CR should be treated is kept with the CR rather than a split between the CR and the metadata.yaml. This will make it easier to maintain.

The question of enumerating fields is, to me, one of the key questions. The competing interests are
* The tool is meant to highlight deviations (risk) between the user's config and the reference. If the user adds in config that is unexpected (ie not in the reference) that is potentially a risk and should be reported to the user for review.
* Reporting many irrelevant diffs creates noise making the tool less valuable. Requiring work by reference maintainers to suppress content may increase the noise.
As a risk assessment tool I think our bias needs to be towards showing the diffs over potentially suppressing a relevant risk. That said, I think your proposal here can help the reference designer add the needed info to suppress more than we are today:
* Support the inverse of the exact-match annotation, something like "unspecified-fields-are-ok". This lets the reference designer annotate an entire CR as "mergable" so that any unspecified fields are automatically OK. Namespaces are good examples where this would be used. I prefer the 'opt-in' form of this annotation rather than exact-match which is an 'opt-out'
* Support field-suppression annotation in the CRs (ie a comment similar to your example) where the designer can omit fields (or a tree if the field is a complex data type) based on jsonpath. As you noted this needs to be per template (CR) and, to me, is easier to maintain along with the reference CR.
Adding support for both of these will provide the tools are in the reference designers hands on how to handle suppression of diffs. They can make the decision on what is important to highlight and what can be suppressed. 
Regarding the burden this adds to the reference designer, I think this is where the high-value work exists in this tool flow. The reference designer can do the analysis/calculus once and all the users benefit from the work. Typically the reference designer is well aware of changes that will come into the CRs release-over-release because they have to account for this in the overall design anyway. Capturing it in the reference is one of the ways way they can communicate that change.

> 
> We could switch the behaviour to off by default instead but you will end up getting false positives with labels, annotations making it annoying to build references. Which is what @pixelsoccupied was facing when he was to building the DU profile.
> 
> While being able to save diffs and passing them in as acceptable goes some what to mitigate this I think would be annoying.
> 
> In some ways it can also make templates more explicit if you really want a field never exist you can set it to null - showing you care about it not being there.
> 
> edit: For now I'll default it off and then we can see how much we need to turn it on, it would be easy to flip the default

nocturnalastro:
@imiller0 That makes sense, I think we have provided at least some of those tools now we can see if more are needed over time.

I think in the future it might be worth extracting extra the comments from the yaml tree to allow more refined usage but ATM I think a custom function while a little cumbersome is good enough to see if its needed.

pkg/compare/compare_test.go
nocturnalastro: I think the flag is still a useful tool for people building references to see the effect of merging and also for people who want to expose any diffs hidden by merging.

docs/reference-config-guide.md
```diff
@@ -105,3 +105,130 @@ spec:
...
+This will do a strategic merge of the template into the manifest which will remove features not mentioned in the template from the diff.
+This can be useful when dealing with annotation or labels which may be of no consiquence to your check.
+Note that this could cover up differences in things you do care about so use it with care.
+This can be configured for a manifest by adding
+`# cluster-compare: allow-undefined-extras=true` in your template
+
+example when comparing the template:
+
+```yaml
+# cluster-compare: allow-undefined-extras=true
```
pixelsoccupied, line 123: Maybe consider a bit more "known" syntax like e.g how we have kernel args? e.g 
`# CLUSTER_COMPARE=allow_undefined_extras=true another_config=true`

With this self-contained config...so have two places to configure stuff? i.e configs inside the CR + metadata.yaml? Should be this be converged?

AlinaSecret: I think going for a yaml syntax can be a good solution:

#cluster-compare start 
#allow-undefined-extras: true
#cluster-compare end

this will allow:
1. easy addition of fields
2. a known syntax with many options (nested values)
3. the code will be more neat because the parsing is a lot easier, and in case there are keys that are not supported in the config an error will be raised automatically ( i will add a comment with more explanations in the relevant section in the code)

AlinaSecret: @pixelsoccupied about what you said about this:
```
With this self-contained config...so now we have two places to configure stuff? i.e configs inside the CR + metadata.yaml? Should be this be converged?
```

the role of the metadata.yaml is to set global reference settings and relations of multiple templates (grouping). this feature adds rules about only a specific template and this rule depends on the structure of the template that is why its better and more easier to maintain ,  when writing\viewing the template its easier to see all settings in the same file, and understand the structure of the template based on these values

pkg/compare/parsing.go
```diff
@@ -179,21 +220,62 @@ func parseDiffConfig(filePath string) (UserConfig, error) {
...
+	yamlTemplate, err := t.Exec(map[string]any{})
 	return yamlTemplate, err
 }
+
+func extractCommentMap(temp *template.Template) (map[string]string, error) {
+	config := make(map[string]string)
+	raw, _ := executeYAMLTemplatRaw(temp, map[string]any{})
+	scanner := bufio.NewScanner(bytes.NewReader(raw))
+	errs := make([]error, 0)
+	for scanner.Scan() {
```
AlinaSecret, line 253: In general we should try to make this function as elegant as possible, see my comment about switching to yaml,

that way this part should only extract the yaml from the comments and load it in to a struct that, 
will automatically make sure that there are no other unsupported keys in the config.

pkg/compare/parsing.go
```diff
@@ -132,25 +136,62 @@ func parseYaml[T any](fsys fs.FS, filePath string, structType *T, fileNotFoundEr
 	return nil
 }
 
-func parseTemplates(templatePaths, functionTemplates []string, fsys fs.FS) ([]*template.Template, error) {
-	var templates []*template.Template
+type ReferenceTemplate struct {
+	*template.Template
+	allowMerge bool
+	config     map[string]string
+}
```
AlinaSecret, line 143: when transferring to yaml you can have a complete defined struct with all of the features supported in the per template config, instead of the config + allow merge 
also will be great for  package documentation

pkg/compare/parsing.go
```diff
@@ -132,25 +136,62 @@ func parseYaml[T any](fsys fs.FS, filePath string, structType *T, fileNotFoundEr
...
+func NewReferenceTemplate(temp *template.Template) (*ReferenceTemplate, error) {
+	rf := &ReferenceTemplate{Template: temp, config: make(map[string]string), allowMerge: true}
+
+	config, err := extractCommentMap(temp)
+	if err != nil {
+		return rf, err
+	}
+
+	rf.config = config
+	v, ok := config["allow-undefined-extras"]
```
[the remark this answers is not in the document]
AlinaSecret: Note: maybe after the change of the syntax to yaml, this will be less relevant. but still leaving it for reference :)

pkg/compare/testdata/CheckNoOverwrite/reference/namespace.yaml
```diff
@@ -0,0 +1,9 @@
+# cluster-compare: allow-undefined-extras=true
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: openshift-storage
+  annotations:
+    workload.openshift.io/allowed: management
+  labels:
+    openshift.io/cluster-monitoring: "true"
```
AlinaSecret, line 9: can you add a comment here that this is the filed that is checked to not be overwrite

pkg/compare/testdata/OnlyMatchPartial/reference/namespace.yaml
```diff
@@ -0,0 +1,9 @@
+# cluster-compare: allow-undefined-extras=true
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: openshift-storage
+  annotations:
+{{ onlyMatch (or .metadata.annotations (dict)) "workload.openshift.io/allowed: management" | indent 4}}
```
AlinaSecret, line 7: Just something to keep in mind i think the usage of these functions is harder to read then using the go tempting syntax implementation, i would include the examples on how to do it with template syntax only .
just something to think about it, I will try it out myself and update you :)  @nocturnalastro (not sure about this and need to check it out, i will update you, just wanted you to know that i am concerned from the usage)

pkg/compare/compare_test.go
```diff
@@ -329,6 +330,47 @@ error code:2`),
...
+			checks: defaultChecks,
+		},
+		{
+			name:   "Check Probes Only Match",
+			mode:   []Mode{DefaultMode},
+			checks: defaultChecks,
+		},
+		{
+			name:     "Show More Flag",
+			mode:     []Mode{DefaultMode},
```
AlinaSecret, line 365: please add a test that includes additional unsupported per-template settings, and make sure an error is raised 
it can be even a typo in  allow-undefined-extras ( eg: allow-undefined-extra )

pkg/compare/compare_test.go
AlinaSecret: totally agree with you that users should not change the reference! 
its even exactly my point, this setting is part of the reference, the reference builder is fully in charge of it.
 the reference builder is in charge complete of the full testing of the reference, meaning if its set to ok to merge then it is ok.
 
 i don't agree that it is compatible to verbosity, by this approach you can take i to the direction of the extreme and say the same about any other rule in the reff, and we dont have a flag for that ( or the example i gave for the fields to ommit, its also possible users would like to view things there that were chose to be supressed by the ref builder)  

but totally i am happy to hear you opinion! continue with the great work! and yeah we can revisit everything, and try to merge in small parts the things we are sure about

AlinaSecret:
@imiller0 @nocturnalastro @pixelsoccupied 

as Ian mentioned :
```
Support field-suppression annotation in the CRs (ie a comment similar to your example) where the designer can omit fields (or a tree if the field is a complex data type) based on jsonpath. As you noted this needs to be per template (CR) and, to me, is easier to maintain along with the reference CR.
```
 also discussed here in a different worse implementation    (https://github.com/openshift/kube-compare/pull/37#discussion_r1646334450) 

i think we should address this in a different issue, for now , for templates that may need this feature the `merge` should be disabled and we can use go template solutions, i will add some clear examples, that can be added to the docs.

and we can open in issue for this change, i think after this PR we will have a good base to add it in

pkg/compare/testdata/OnlyMatchPartial/reference/namespace.yaml
nocturnalastro: The main issue I found is that when the template is rendered with the empty map I was getting a issues so had to set a default.

> harder to read then using the go tempting syntax implementation

Do you enumerating through as a loop and setting them to null?

docs/reference-config-guide.md
AlinaSecret: update:
instead of 
```
#cluster-compare start
#allow-undefined-extras: true
#cluster-compare end
```
you can do:

```
#templateConfig:
#    allow-undefined-extras: true
```

pixelsoccupied: > you can do:

we should stick to a single line config with a well-know key to simplify the custom config syntax. Otherwise to be complete we need a full blown AST. Check out of discussion here https://github.com/go-yaml/yaml/issues/132/ 

But my concern is not the syntax itself but if we can use something that's already established/well-known.


Thinking about this a bit more...can we rely on go-template for this? 

e.g 
```
{{ template "AllowUndefinedExtras"  }}
apiVersion: v1
kind: Namespace
metadata:
  name: openshift-storage
  annotations:
    workload.openshift.io/allowed: management
  labels:
    openshift.io/cluster-monitoring: "true"
```

And then we define `AllowUndefinedExtras` in go which can change modify the config?

nocturnalastro: This seems like a good idea then we can change it later to whatever we want. Perhaps we could use a function which returns something instead of a template so it looks like 
```
{{ ClusterCompareConfigAllowUndefinedFields  }}
apiVersion: v1
kind: Namespace
metadata:
  name: openshift-storage
  annotations:
    workload.openshift.io/allowed: management
  labels:
    openshift.io/cluster-monitoring: "true"
```
that way we don't run into any templates that other people may define.

nocturnalastro: I would also like to support something like this

```yaml
apiVersion: v1
kind: Namespace
metadata:
  name: openshift-storage
  annotations: # cluster-compare: allow-undefined-extras=true
      workload.openshift.io/allowed: management
```

nocturnalastro: We after a discussion decided to put the config into the metadata file.

docs/reference-config-guide.md
```diff
@@ -105,3 +105,130 @@ spec:
     tier: frontend
     {{- end }}
 ```
+
+# Per-template configuration
+
```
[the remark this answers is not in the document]
AlinaSecret: marking as Not relevant any more, too make sure all comments are resolved

pkg/compare/compare_test.go
```diff
@@ -92,6 +92,10 @@ func checkFile(t *testing.T, fileName, value string) {
 	if err != nil {
 		t.Fatalf("test %s failed reading .golden file: %s", fileName, err)
 	}
+	fmt.Println(string(expected))
```
AlinaSecret, line 95: i think its a little to verbose, why do you think its necessary?

pkg/compare/compare.go
```diff
@@ -498,8 +502,15 @@ func (obj InfoObject) Live() runtime.Object {
 
 // Merged Returns the Injected Reference Version of the Resource
 func (obj InfoObject) Merged() (runtime.Object, error) {
+	var err error
+	if obj.allowMerge {
+		obj.injectedObjFromTemplate, err = mergeManifests(obj.injectedObjFromTemplate, obj.clusterObj)
+		if err != nil {
+			klog.Errorf("failed to properly merge the manifests for %s some diff may be incorrect: %s", apiKindNamespaceName(obj.clusterObj), err)
```
AlinaSecret, line 509: why is the error raised with klog instead of just returning an error, i don't see an reason to treat the error different then the rest

pkg/compare/parsing.go
```diff
@@ -35,14 +35,28 @@ const (
 )
 
 type Component struct {
-	Name              string        `json:"name"`
-	Type              ComponentType `json:"type,omitempty"`
-	RequiredTemplates []string      `json:"requiredTemplates,omitempty"`
-	OptionalTemplates []string      `json:"optionalTemplates,omitempty"`
+	Name string        `json:"name"`
+	Type ComponentType `json:"type,omitempty"`
+	// Config 			  ReferenceTemplateConfig         `json:"config,omitempty"` // TODO figure out how to make this work
```
AlinaSecret, line 40: i guess this was left here not on purpose (right?)

pkg/compare/parsing.go
```diff
@@ -179,21 +192,27 @@ func parseDiffConfig(filePath string) (UserConfig, error) {
 
 const noValue = "<no value>"
 
-func executeYAMLTemplate(temp *template.Template, params map[string]any) (*unstructured.Unstructured, error) {
+func executeYAMLTemplatRaw(temp *template.Template, params map[string]any) ([]byte, error) {
```
AlinaSecret, line 195: i think there is no need for this function anymore,  all of the usages are with the executeYAMLTemplate function

pkg/compare/testdata/CheckProbesMerge/reference/pod.yaml
```diff
@@ -0,0 +1,22 @@
+# cluster-compare: allow-undefined-extras=true
```
AlinaSecret, line 1: can you remove this from all of the test files?

pkg/compare/compare_test.go
```diff
@@ -329,6 +333,21 @@ error code:2`),
...
+			mode:   []Mode{DefaultMode},
+			checks: defaultChecks,
+		},
+		{
+			name:   "Check No Overwrite",
+			mode:   []Mode{DefaultMode},
+			checks: defaultChecks,
+		},
+		{
+			name:   "Check Probes Merge",
```
AlinaSecret, line 347: what does this test check?

pkg/compare/compare_test.go
```diff
@@ -329,6 +333,21 @@ error code:2`),
 			outputFormat: Json,
 			checks:       defaultChecks,
 		},
+		{
+			name:   "Allow Undefined Extras",
+			mode:   []Mode{DefaultMode},
+			checks: defaultChecks,
+		},
+		{
+			name:   "Check No Overwrite",
```
AlinaSecret, line 342: can you change the name of **all of the tests** to be informative - should include that they test the "Allow Undefined Extras"\Strategic Merge and also what is the exact test case, example:

Strategic Merge cluster CR not Overwrites reference 

so it will be easy to understand by the name of the test\dir what exactly is this testcase

pkg/compare/parsing.go
```diff
@@ -35,14 +35,28 @@ const (
...
+	OptionalTemplates []*ReferenceTemplate `json:"optionalTemplates,omitempty"`
+}
+type ReferenceTemplateConfig struct {
+	AllowMerge bool `json:"allow-undefined-extras,omitempty"`
+}
+
+type ReferenceTemplate struct {
+	*template.Template
+	Path   string                  `json:"path"`
+	Config ReferenceTemplateConfig `json:"config,omitempty"`
```
AlinaSecret, line 51: wondering why you want to leave the config section? i think it doesn't have some extra value and may be more annoying, like in Component the type is under component not under config    
anyway i don't have a string opinion so whatever you think is best

AlinaSecret (commented):
Great work, i think were getting close to finishing this one :)

pkg/compare/compare_test.go
AlinaSecret: marking as Not relevant any more, too make sure all comments are resolved

pkg/compare/testdata/OnlyMatchPartial/reference/namespace.yaml
AlinaSecret: was moved to different PR

pkg/compare/parsing.go
AlinaSecret: marking as Not relevant any more, too make sure all comments are resolved

pkg/compare/funcmap.go
```diff
@@ -150,3 +153,45 @@ func fromJSONArray(str string) []any {
...
 	return a
 }
+
+// onlyMatch does the inverse of the merge function
+//
+// Can be useful when using allow-undefined-extras=true
+// as the merges will cancel out allowing you to be more
+// specific for a block.
+// If an error occurs it will return error message string
+func onlyMatch(data any, matchObj string) string {
```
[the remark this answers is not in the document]
AlinaSecret: separated

pkg/compare/parsing.go
nocturnalastro: I think its a good idea for config to have its own type, I think it will make things like allowing an ansible style hierarchical config easier. We can always remove it later if it turns out to be a pain.

pkg/compare/compare_test.go
nocturnalastro: I don't think this test is needed that was sanity check for me to check the merge would handle strange cases correctly.

pkg/compare/compare_test.go
nocturnalastro: Oh sorry that was to help me when debugging. Easier to read the println than to look at the string

pkg/compare/compare.go
nocturnalastro: We can still do the diff of the un-merged manifest, my reasoning was that an overly verbose diff is better than nothing.

pkg/compare/compare_test.go
```diff
@@ -329,6 +329,16 @@ error code:2`),
 			outputFormat: Json,
 			checks:       defaultChecks,
 		},
+		{
+			name:   "Allow Undefined Extras",
```
AlinaSecret, line 333: change name to Check Merging ...... (not sure what is exactly tested here, maybe: only specified diffs are shown, in a different phrasing )  

also because we changed the name of the flag this should change to

pkg/compare/compare.go
```diff
@@ -433,27 +434,28 @@ func (o *Options) Run() error {
 		return fmt.Errorf("failed to collect resources: %w", err)
 	}
 	r.IgnoreErrors(func(err error) bool {
-		return containOnly(err, []error{MultipleMatches{}, UnknownMatch{}})
+		return containOnly(err, []error{MultipleMatches{}, UnknownMatch{}, MergeError{}})
```
AlinaSecret, line 437: you don't have to put  the mergeerror inside the containOny because the error isn't joined, only the corealtor returns a  joined errors that need this behavior, but i guess it should work like this also .
so up to you, but just for notice

pixelsoccupied:
Amazing stuff! Looking forward to updating the reference with this! 

/lgtm
/assign @AlinaSecret

pkg/compare/compare.go
nocturnalastro: It has a clause for non-joined errors, I thought it would be good re-use that as to not make more branches and vars than needed.

AlinaSecret:
/lgtm
/approve
Thanks @nocturnalastro know it took more time then the usual and may been exhausting , so i think its a big win to get this one merged  :1st_place_medal:  !!!!  Appreciate your effort very much :)

Issue #31, opened 2024-06-11
Custom go template function throwing an error 


pixelsoccupied:
```
error: template: odfNS.yaml:10:16: executing "storage/odf-external/odfNS.yaml" at <{{template "hello"}}>: template "hello" not defined
```

```
{{define "hello"}}
    hello: world
{{end}}
apiVersion: v1
kind: Namespace
metadata:
  name: openshift-storage
  annotations:
    workload.openshift.io/allowed: management
    {{ template "hello" }}
  labels:
    openshift.io/cluster-monitoring: "true"
```

It should render instead of the error
```
apiVersion: v1
kind: Namespace
metadata:
  name: openshift-storage
  annotations:
    workload.openshift.io/allowed: management
    hello: world
  labels:
    openshift.io/cluster-monitoring: "true"
```

PR #40, opened 2024-06-20
Fix use of template functions declared in the same file

AlinaSecret:
When using template functions that are declared in the same file as
 the templates, an error would be returned specifying that the function
 is not found. This commit fixes this behavior and also adds a test case 
that ensures that the bug has been fixed.

### DIFF
--- a/pkg/compare/compare_test.go
+++ b/pkg/compare/compare_test.go
@@ -404,6 +404,11 @@ error code:2`),
 			name:   "Invalid Resources Are Skipped",
 			checks: defaultChecks,
 		},
+		{
+			name:   "Ref Contains Templates With Function Templates In Same File",
+			mode:   []Mode{DefaultMode},
+			checks: defaultChecks,
+		},
 	}
 	tf := cmdtesting.NewTestFactory()
 	testFlags := flag.NewFlagSet("test", flag.ContinueOnError)
--- a/pkg/compare/parsing.go
+++ b/pkg/compare/parsing.go
@@ -154,7 +154,7 @@ func parseTemplates(templateReference []*ReferenceTemplate, functionTemplates []
 			continue
 		}
 		// recreate template with new name that includes path from reference root:
-		parsedTemp, _ = template.New(temp.Path).Funcs(FuncMap()).AddParseTree(temp.Path, parsedTemp.Tree)
+		parsedTemp, _ = parsedTemp.New(temp.Path).AddParseTree(temp.Path, parsedTemp.Lookup(path.Base(temp.Path)).Tree)
 		if len(functionTemplates) > 0 {
 			parsedTemp, err = parsedTemp.ParseFS(fsys, functionTemplates...)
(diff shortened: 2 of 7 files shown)


AlinaSecret:
@pixelsoccupied @nocturnalastro

AlinaSecret:
close #31

pkg/compare/parsing.go
```diff
@@ -142,7 +142,7 @@ func parseTemplates(templatePaths, functionTemplates []string, fsys fs.FS) ([]*t
 			continue
 		}
 		// recreate template with new name that includes path from reference root:
-		parsedTemp, _ = template.New(temp).Funcs(FuncMap()).AddParseTree(temp, parsedTemp.Tree)
+		parsedTemp, _ = parsedTemp.New(temp).Funcs(FuncMap()).AddParseTree(temp, parsedTemp.Lookup(path.Base(temp)).Tree)
```
pixelsoccupied, line 145: Curious why we are recreating template? 

If we init just one time (L139) then we can simply do `parsedTemp, _ = parsedTemp.Funcs(FuncMap()).AddParseTree(temp, parsedTemp.Tree)` ?

nocturnalastro: Also I think if we don't recreate the template I don't think we need to re-add the the funcmap

pkg/compare/compare_test.go
```diff
@@ -329,6 +329,12 @@ error code:2`),
 			outputFormat: Json,
 			checks:       defaultChecks,
 		},
+		{
+			name:         "Ref Contains Templates With Function Templates In Same File",
+			mode:         []Mode{DefaultMode},
+			outputFormat: Json,
```
nocturnalastro, line 335: The test would be easier to check by eye if it wasn't json output :)

pkg/compare/parsing.go
AlinaSecret: we recreate the template because we need to change its name, for future usage in the code.
the name gets automatically choose as the name of the file, that is not good for us in case of templates included in sub-directories, there is now way to change only the name, then we recreate it with the name we want including the relative path from the metadata directory 


maybe after we merge some of @nocturnalastro PRs we can add a name attribute to the wrapped template instead, need to still think about it. but that is another story.

pkg/compare/testdata/RefContainsTemplatesWithFunctionTemplatesInSameFile/reference/cm.yaml
```diff
@@ -0,0 +1,11 @@
+{{- define "addBlob" -}}
+  function was called successfully from different file
```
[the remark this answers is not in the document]
AlinaSecret: yes! you are right pedantic comments are good! nice eye

nocturnalastro:
Lets get this fix in we can worry about moving stuff into the object later :) 

/lgtm
/assign @pixelsoccupied

Issue #20, opened 2024-05-28
Document how to use of go-template custom func or external library (sprig) functions


pixelsoccupied:
https://masterminds.github.io/sprig/

nocturnalastro:
Could move the custom functions into there own package and then link from the docs to the page on https://pkg.go.dev

pixelsoccupied:
@nocturnalastro sorry just update the title to make it a bit more clear. 

We have imported Sprig as one of the dependencies which allows us to funcs like `has` to test if an element is persent etc....was thinking if it can be documented 

Also we can have custom func written in go-template contained with a CR (a.k.a template function) to validate stuff further

pixelsoccupied:
> Could move the custom functions into there own package and then link from the docs to the page on https://pkg.go.dev/
But that's a good idea!

PR #41, opened 2024-06-21
Move funcmap and add template functions docs

nocturnalastro:
#20 

Note the url https://pkg.go.dev/github.com/openshift/kube-compare/pkg/funcmap#pkg-functions wont be live until we merge it but https://pkg.go.dev/github.com/openshift/kube-compare/pkg/compare#pkg-functions to see how it would look - although currently the to/from helper functions are not visible as they are currently private

### DIFF
--- a/docs/templating-functions.md
+++ b/docs/templating-functions.md
@@ -1,0 +1,6 @@
+# Templating functions
+
+On top of golangs normal templating we provide some functions to make writitng references easier.
+The (sprig functions)[http://masterminds.github.io/sprig/] are included along with some other functions that can be found in the (subpackages pkg.go.dev page)[https://pkg.go.dev/github.com/openshift/kube-compare/pkg/funcmap#pkg-functions]
+
+If you want to custom functions you can define them as a templates and include them a paths under `templateFunctionFiles` at the root of your reference `metadata.yml`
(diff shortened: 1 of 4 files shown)


docs/templating-functions.md
```diff
@@ -0,0 +1,6 @@
+# Templating functions
+
+On top of golangs normal templating we provide some functions to make writitng references easier.
+The (sprig functions)[http://masterminds.github.io/sprig/] are included along with some other functions that can be found in the (subpackages pkg.go.dev page)[https://pkg.go.dev/github.com/openshift/kube-compare/pkg/funcmap#pkg-functions]
+
+If you want to custom functions you can define them as a templates and include them a paths under `templateFunctionFiles` at the root of your reference `metadata.yml`
```
pixelsoccupied, line 6: Maybe an example or two (sprig && custom) would be nice!

nocturnalastro: Not sure what you mean here? 
Do you mean you want me to list a few or show a template with them in?

pixelsoccupied: Yep that's it. Show a template with them...so like as a user I would like tips on how to incorporate spring/template  funcs in my existing references! Also not sure what `templateFunctionFiles` is...a custom dir name? (but including an example should clear it up)

btw metadata.yml should be metadata.`yaml`? The file ext is somewhat hardcoded right?

PR #43, opened 2024-06-23
CNF-12627: Change kubectl prefix in help message

natifridman:
Change kubectl prefix in help message to show the correct command for standalone


AlinaSecret:
@natifridman Thanks for the PR !!!! @

PR #45, opened 2024-06-24
CNF-11871: Add abiltiy to save patches and load them as acceptable diffs

nocturnalastro:


pkg/compare/compare.go
```diff
@@ -173,6 +177,10 @@ func NewCmd(f kcmdutil.Factory, streams genericiooptions.IOStreams) *cobra.Comma
 			"In local mode will try to match all resources passed to the command")
 
 	cmd.Flags().StringVarP(&options.OutputFormat, "output", "o", "", fmt.Sprintf(`Output format. One of: (%s)`, strings.Join(OutputFormats, ", ")))
+
+	cmd.Flags().StringVarP(&options.PatchesSavePath, "patches-save-path", "p", "", "Path file where patches should be saved.")
```
pixelsoccupied, line 181: Readme with an exmaple use case please! 

So this cover what Dave was asking yeah? Giving customer to silence diffs overtime?

pkg/compare/compare.go
```diff
@@ -518,11 +583,41 @@ func (obj InfoObject) Name() string {
 	return slug.Make(apiKindNamespaceName(obj.clusterObj))
 }
 
+func (obj InfoObject) getPatch() (string, error) {
+	localRuntime, err := obj.Merged()
```
AlinaSecret, line 587: i am not sure if the patch should created from the version that already contains the patched version (like still not  sure about the case if there is a load and save patch at the same time  what will each of them include)

i would of expect here to use the object only with the ommited fields (without the load patch)

pkg/compare/compare.go
```diff
@@ -624,3 +719,20 @@ func (o Output) Print(format string, out io.Writer) (int, error) {
 	}
 	return n, nil
 }
+
+func (o Output) PrintPatches(out io.Writer) (int, error) {
```
AlinaSecret, line 723: i think its better to create a separate object for the patches, where has both the load function and the print functions. no need to use the ouptput that contains final output of the tool or the diff sum. 

also because the patch file extends the api of the command, i think its better that is format is located in one   place. 
the code is a little to much divided between multiple places, that can make it hard to change and its harder to review.  

so i suggest on creating a struct called PatchFile that has list/map of the patchsave object (that can be just patch)
the patch file object can contain the print function (i would call it export instead or save), and also a load function. even put all of the patch functions in a different file

pkg/compare/compare.go
```diff
@@ -455,26 +472,49 @@ func (o *Options) Run() error {
 			clusterObj:              &clusterCR,
 			FieldsToOmit:            o.ref.FieldsToOmit,
 		}
+		crName := apiKindNamespaceName(&clusterCR)
+		if patches, ok := o.patches[crName]; ok {
```
AlinaSecret, line 476: why are the patches mapped by a specific CR, instead of the corresponding reference.
it can be more stable to changes in the CRS name-namespace-apiversion-kind. and its logical that you would like to supresss a diff because of a field in the ref.

pkg/compare/compare_test.go
```diff
@@ -149,12 +146,52 @@ type Test struct {
...
+	defer func() {
+		_ = recover()
+		test.checks.Out.check(t, *test, mode, removeInconsistentInfo(t, out.String()))
+	}()
+	cmd.Run(cmd, []string{})
+}
+
 // TestCompareRun ensures that Run command calls the right actions
 // and returns the expected error.
 func TestCompareRun(t *testing.T) {
```
AlinaSecret, line 197: please add a test that gets as input the patch file (that is saved in the tests directory)  and tests that the diff is supressed, and same for the save option. 
because the patch file is part of the api it should be fully tested  as a separate file 
also it will be super helpful for the review  too see examples of the patch file

AlinaSecret (commented):
I think its enough for this review, still have some more comments about the tests but they may change according to the other comments so didn't add them for now

pkg/compare/compare.go
nocturnalastro: Yeah good point now that the other PR is merged I'll have to update this one :)

nocturnalastro:
/hold 
while discuss the design

nocturnalastro:
@AlinaSecret @pixelsoccupied Closing this one for now. I'll open up a new once once we've agreed on the design with the new implementation.

PR #49, opened 2024-07-01
CNF-13287: Hide empty diffs by default

nocturnalastro:
Before
```
$ go run cmd/kubectl-cluster_compare.go -r pkg/compare/testdata/RequiredResourcesOfOptionalComponentAreReportedMissingIfAtLeastOneOfResourcesInGroupIsIncluded/reference -f pkg/compare/testdata/RequiredResourcesOfOptionalComponentAreReportedMissingIfAtLeastOneOfResourcesInGroupIsIncluded/resources

**********************************

Cluster CR: v1_Namespace_kubernetes-dashboard
Reference File: ns.yaml
Diff Output: None

**********************************

Summary
CRs with diffs: 0
CRs in reference missing from the cluster: 1
ExamplePart1:
  Dashboard1:
  - cm.yaml
No CRs are unmatched to reference CRs
exit status 1

```

After
```
$ go run cmd/kubectl-cluster_compare.go -r pkg/compare/testdata/RequiredResourcesOfOptionalComponentAreReportedMissingIfAtLeastOneOfResourcesInGroupIsIncluded/reference -f pkg/compare/testdata/RequiredResourcesOfOptionalComponentAreReportedMissingIfAtLeastOneOfResourcesInGroupIsIncluded/resources
Summary
CRs with diffs: 0
CRs in reference missing from the cluster: 1
ExamplePart1:
  Dashboard1:
  - cm.yaml
No CRs are unmatched to reference CRs
exit status 1
```
with `-v`
```
$ go run cmd/kubectl-cluster_compare.go -r pkg/compare/testdata/RequiredResourcesOfOptionalComponentAreReportedMissingIfAtLeastOneOfResourcesInGroupIsIncluded/reference -f pkg/compare/testdata/RequiredResourcesOfOptionalComponentAreReportedMissingIfAtLeastOneOfResourcesInGroupIsIncluded/resources -v
**********************************

Cluster CR: v1_Namespace_kubernetes-dashboard
Reference File: ns.yaml
Diff Output: None

**********************************

Summary
CRs with diffs: 0
CRs in reference missing from the cluster: 1
ExamplePart1:
  Dashboard1:
  - cm.yaml
No CRs are unmatched to reference CRs
exit status 1
```


pkg/compare/compare_test.go
```diff
@@ -136,6 +141,13 @@ type Checks struct {
 	Err Check
 }
 
+func (c Checks) withPrefixedSuffix(suffixPrefix string) Checks {
```
AlinaSecret, line 144: can you add a comment explaining that a it changes the suffix of the **tests expected output files**.
and add that this function can allow having multiple tests with the same ref+resources but with different arguments, without the need to create separate test dir.  

in general just want it to be more easier for new people to  understand.
I think this addition is great and we also can combine some of the test directories that are duplicated due to different arguments (for example i think the yaml\json output)

AlinaSecret (commented):
looks good, minor comment

pkg/compare/compare_test.go
nocturnalastro: Yeah I though it would be useful for things like that :)

nocturnalastro: Done :)

natifridman:
Maybe it's worth adding a before/after example to this PR description?

nocturnalastro:
> Maybe it's worth adding a before/after example to this PR description?

Done just picked a random piece of test data :)

PR #50, opened 2024-07-01
Fix tests when using MacOS

nocturnalastro:


natifridman (approved):
/lgtm
Works on my Mac M3

AlinaSecret:
@nocturnalastro in the end the yaml check is more elegant then i expected :) 
looks good, Thanks!!! 
\lgtm

pixelsoccupied:
/lgtm 
/approve

Looks like it's more generic that just macos specific! Thanks!

PR #51, opened 2024-07-01
Allow every part to be quoted

nocturnalastro:


pkg/compare/parsing.go
```diff
@@ -101,24 +178,71 @@ func (r *Reference) getMissingCRs(matchedTemplates map[string]bool) (map[string]
...
+	"metadata.uid",
+	"metadata.generateName",
+	"metadata.creationTimestamp",
+	"metadata.finalizers",
+	`"kubectl.kubernetes.io/last-applied-configuration"`,
+	`metadata.annotations."kubectl.kubernetes.io/last-applied-configuration"`,
+	"status",
+}
+
+type Path struct {
```
AlinaSecret, line 195: This part is a declaration of a new "language",  like jsonPath, just that for now the implementation does not include many features so it may look not that big of a deal.

I don't think its a good idea to implement a custom syntax for this path,    its better to use some thing that already exits (we can look in to it together ).

Also if we do choose implementing are own syntax, this should be placed in a different package and well documented. also i think the implementation should be changed to be more extendable, for example if we are going to support regex this will need refactoring, and will make it more complex to understand. I think in this case there are neater ways to define parsed languages in go lang (quick search: https://github.com/alecthomas/participle probaly there are better then that also)
    
I do strongly think its smart to look for a known approach that  is already implemented, and invest the time in finding one. also possible to take a known solution and  remove features from it or constraint if its to complex.
  I think this also connect to are discussion when we wanted to add comments to the templates, where also the outcome was not to create new formats.

pkg/compare/parsing.go
```diff
@@ -34,14 +37,72 @@ const (
...
+		for _, path := range pathsArray {
+			p, err := NewPath(path)
+			if err != nil {
+				klog.Errorf("skipping path: %s", err)
+				continue
+			}
+			processedPaths = append(processedPaths, p)
+		}
+		if len(processedPaths) == 0 {
+			klog.Errorf("skipping key: no paths in key %s", key)
```
AlinaSecret, line 89: lets return errors instead of using klog,
if there is a problem it should not skip, this isnt a dynamic problem, its a problem in the ref structure, this can not be a vaild ref, it will make this pop for the user each time. 
(and are you sure it skips?  i remember the klog.errorf exits, anyway i think  it should not skip ) 
if you want to return all the errors from this function  i think its better to use error.join and then return it so the user gets all the errors at the same run of the command.

AlinaSecret (changes requested):
Left some comments,  liked the option that you can reference the fields to ommit for specific CRs 
still have not finished  doing a full pass (but i am posting these in the mean time )

pkg/compare/parsing.go
AlinaSecret: @pixelsoccupied what do you think?

AlinaSecret: update: take a look at the first answer here:
https://stackoverflow.com/questions/47489745/splitting-a-string-at-space-except-inside-quotation-marks
i guess also an option is to simplify it to be a like csv with a dot separator untill think about the regexes

pkg/compare/parsing.go
nocturnalastro: I'll take a look but yes thats fair it should fail.

pkg/compare/parsing.go
```diff
@@ -34,14 +37,72 @@ const (
...
+	fieldsToOmitDefaultNotFound    = `fieldsToOmit's defaultKey "%s" not found in items`
+	fieldsToOmitRefsNotFound       = `skipping fieldsToOmitRefs entry: "%s" not found it fieldsToOmit Items`
+)
+
+type FieldsToOmitConfig struct {
+	DefaultKey string `json:"defaultKey,omitempty"`
+}
+
+type FieldsToOmit struct {
+	Config         FieldsToOmitConfig  `json:"config,omitempty"`
```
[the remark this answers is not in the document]
nocturnalastro: This was a suggestion from @pixelsoccupied, I'm not fussed either way. Theres pros/cons to both.

pkg/compare/parsing.go
nocturnalastro: If it starts getting more complex than the regex case (e.g. logic or more extensive operators) perhaps we should go down the route of a properly defined language with a parser and lexer but ATM its overkill. ATM its convenient way of describing a list of segments (even with regex) and I think it should stay that way.

I went down the path of using a jq lib but due to the complexity of jq it wasn't easy to do the back tracking, to clean up the empty structures. 
Thinking about it perhaps we could instead of deleting the item, we could replace it with a marker that we can then search for and delete, then do the back tracking. I can try this if you would like.

Its not very hard to extend it to regex can just denote regex with something like backticks. https://github.com/nocturnalastro/kube-compare/pull/2/files#diff-a82c27d824a8a061c8c740353caf25344e333a1a4d77c6c2e2d57824a5cb31b1R209
I actually initially wrote it with the regex case in but thought it needed some refinement so I split it off. 

There is other ways to approach parsing the line though that might be clearer I'll give some of them a go for comparison.

pkg/compare/parsing.go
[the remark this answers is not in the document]
pixelsoccupied: I'm ok with either...if want to drop config maybe update "defaultKey" to be a bit more verbose e.g "defaultOmitRule" ?

pkg/compare/parsing.go
nocturnalastro: Given that we only need prefix and not full regex support we decided to go with csv reader to split and then just a flag for isPrefix

AlinaSecret (commented):
before i finish the review  can you please tidy up the PR as specified in slack .
(part of it is based on: https://github.com/kubernetes/community/blob/master/contributors/guide/pull-requests.md#squashing)

nocturnalastro:
> before i finish the review can you please tidy up the PR as specified in slack . (part of it is based on: https://github.com/kubernetes/community/blob/master/contributors/guide/pull-requests.md#squashing)

Sure but in general its much easier to clean up the commits once we agree the code is fine.

pkg/compare/parsing.go
```diff
@@ -50,6 +97,22 @@ type ReferenceTemplate struct {
 	Config ReferenceTemplateConfig `json:"config,omitempty"`
 }
 
+func (rf ReferenceTemplate) FeildsToOmit(fieldsToOmit FieldsToOmit) []*Path {
```
AlinaSecret, line 100: i think that this function should be called additionally or at least only the validation part before running the diff. 
if there is a a ref not found in the  fields to omit its a problem in the reference, that always will occur   and no need for the user no see it on each run. I think its good to enforce ref creators to create a valid ref. so users will not get unwanted warnings that can be confusing. 

i think maybe a validation function for the ref can be a good place to add it, and call the ReferenceTemplate.validate in the complete function (with all the other validation)  or maybe  put it in the getReference func

AlinaSecret, line 100: small typo in function name :)

pkg/compare/testdata/DiffinCustomOmittedFieldsIsntShownLeadingDot/reference/metadata.yaml
```diff
@@ -0,0 +1,15 @@
...
+      - name: Dashboard
+        type: Required
+        requiredTemplates:
+          - path: deploymentMetrics.yaml
+
+fieldsToOmit:
+  defaultOmitRef: deployment
+  items:
+    deployment:
+      - pathToKey: .spec.selector.matchLabels.k8s-app
```
AlinaSecret, line 13: curios, why did you decide to support a leading dot?

pkg/compare/parsing.go
```diff
@@ -101,24 +164,45 @@ func (r *Reference) getMissingCRs(matchedTemplates map[string]bool) (map[string]
...
+	{PathToKey: "metadata.uid"},
+	{PathToKey: "metadata.generateName"},
+	{PathToKey: "metadata.creationTimestamp"},
+	{PathToKey: "metadata.finalizers"},
+	{PathToKey: `"kubectl.kubernetes.io/last-applied-configuration"`},
+	{PathToKey: `metadata.annotations."kubectl.kubernetes.io/last-applied-configuration"`},
+	{PathToKey: "status"},
+}
+
+type Path struct {
```
AlinaSecret, line 181: Can you change the name to be more clear, and self explanatory to the type of the path (so its not confused with os path for example)

AlinaSecret (commented):
cool, i think its getting super close to be merged  :)

pkg/compare/testdata/DiffinCustomOmittedFieldsIsntShownLeadingDot/reference/metadata.yaml
nocturnalastro: some of the similar syntaxes have a leading dot. And its really easy to just ignore it. Didn't see the point in not :)

PR #55, opened 2024-07-07
CNF-13524: Add Dockerfile

natifridman:
- Add Dockerfile
- Modify makefile to support cross-platform build

Building the image
```bash
make image-build 
```

Getting the binaries
```bash
➜ podman run --platform linux/amd64 -it --entrypoint /bin/bash kube-compare
[root@f1b611b07d36 /]# ls
afs  bin  boot	dev  etc  home	lib  lib64  lost+found	media  mnt  opt  proc  root  run  sbin	srv  sys  tmp  usr  var
[root@f1b611b07d36 /]# ls /usr/share/openshift
LICENSE  linux_amd64  linux_arm64  linux_ppc64le  linux_s390x  mac  mac_arm64  windows
```

Based on [oc cli-artifacts](https://github.com/openshift/oc/blob/master/images/cli-artifacts/Dockerfile.rhel) image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,7 +6,7 @@ COPY . .
 RUN go work vendor
 RUN make cross-build --warn-undefined-variables
 
-FROM registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.22-builder-multi-openshift-4.17 AS builder-rhel-9
+FROM registry.ci.openshift.org/ocp/builder:rhel-9-golang-1.22-builder-multi-openshift-4.17 AS builder-rhel-9
 WORKDIR /go/src/github.com/openshift/kube-compare
 COPY . .
 RUN go work vendor
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,35 @@
+# This Dockerfile builds an image containing the Mac and Windows version of kube-compare
+# layered on top of the Linux cli image.
+FROM registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.22-builder-multi-openshift-4.17 AS builder-rhel-8
+WORKDIR /go/src/github.com/openshift/kube-compare
+COPY . .
+RUN go work vendor
+RUN make cross-build --warn-undefined-variables
+
+FROM registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.22-builder-multi-openshift-4.17 AS builder-rhel-9
+WORKDIR /go/src/github.com/openshift/kube-compare
+COPY . .
+RUN go work vendor
+RUN make cross-build --warn-undefined-variables
+
+FROM --platform=linux/amd64 registry.ci.openshift.org/ocp/4.17:cli
+
+COPY --from=builder-rhel-9 /go/src/github.com/openshift/kube-compare/_output/bin/ /usr/share/openshift/
+
+RUN cd /usr/share/openshift && \
+    ln -sf /usr/share/openshift/linux_amd64/kube-compare /usr/bin/kube-compare && \
+    mv windows_amd64 windows && \
+    mv darwin_amd64 mac && \
+    mv darwin_arm64 mac_arm64
+
+COPY --from=builder-rhel-8 /go/src/github.com/openshift/kube-compare/_output/bin/linux_amd64/kubectl-cluster_compare /usr/share/openshift/linux_amd64/kube-compare.rhel8
+COPY --from=builder-rhel-8 /go/src/github.com/openshift/kube-compare/_output/bin/linux_arm64/kubectl-cluster_compare /usr/share/openshift/linux_arm64/kube-compare.rhel8
+COPY --from=builder-rhel-8 /go/src/github.com/openshift/kube-compare/_output/bin/linux_ppc64le/kubectl-cluster_compare /usr/share/openshift/linux_ppc64le/kube-compare.rhel8
+COPY --from=builder-rhel-8 /go/src/github.com/openshift/kube-compare/_output/bin/linux_s390x/kubectl-cluster_compare /usr/share/openshift/linux_s390x/kube-compare.rhel8
+
+COPY --from=builder-rhel-9 /go/src/github.com/openshift/kube-compare/_output/bin/linux_amd64/kubectl-cluster_compare /usr/share/openshift/linux_amd64/kube-compare.rhel9
+COPY --from=builder-rhel-9 /go/src/github.com/openshift/kube-compare/_output/bin/linux_arm64/kubectl-cluster_compare /usr/share/openshift/linux_arm64/kube-compare.rhel9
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -5,10 +5,24 @@ IMAGE_NAME=kube-compare
...
+
+OUTPUT_DIR :=_output
+GO_BUILD_BINDIR ?=$(OUTPUT_DIR)/bin
+CROSS_BUILD_BINDIR ?=$(OUTPUT_DIR)/bin
 
 .PHONY: build
 build:
-	go build ./cmd/kubectl-cluster_compare.go
+	mkdir -p $(GO_BUILD_BINDIR)
+	GOOS=$(GOOS) GOARCH=$(GOARCH) go build $(GO_BUILD_FLAGS) -o $(GO_BUILD_BINDIR)/kubectl-cluster_compare ./cmd/kubectl-cluster_compare.go
```
pixelsoccupied, line 25: Looks good! Do you mind adding a one of two lines with some instruction on how to run this and maybe the what output is? 

Also we are missing a build command for the dockerfile in thie PR? At least example command to build and copy out the binary would nice...and we can update the makefile in the next PR if you would like

natifridman: Good point, I added the image build to the makefile and a comment about the arch types.
Let me know if you think we should add more info there.

Dockerfile
```diff
@@ -0,0 +1,35 @@
+# This Dockerfile builds an image containing the Mac and Windows version of kube-compare
+# layered on top of the Linux cli image.
+FROM registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.22-builder-multi-openshift-4.17 AS builder-rhel-8
+WORKDIR /go/src/github.com/openshift/kube-compare
+COPY . .
+RUN go work vendor
+RUN make cross-build --warn-undefined-variables
+
+FROM registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.22-builder-multi-openshift-4.17 AS builder-rhel-9
```
pixelsoccupied, line 9: `rhel-8-golang-1.22-builder-multi-openshift-4.17` should be rhel9 right?

natifridman: oops, good catch

Makefile
pixelsoccupied: I guess the bit about "how to copy the binary and save it locally" part in a doc? Maybe something like this (should be the same of podman right? )

```
docker create --name kc kc:v1
docker cp kc:/usr/share/openshift/linux_amd64/kube-compare.rhel9 /local-dir/kube-compare.rhel9
docker rm -f kc
```

But that's probably something we can include in the next PR when we link the upcoming container repo with CI...up to you?

natifridman: Added a basic doc explaining this. I think we adjust it in a different PR if needed

docs/image-build.md
```diff
@@ -0,0 +1,24 @@
...
+
+## Build container
+
+```bash
+make image-build
+```
+
+## Copy binary locally
+
+- One option it to build the image locally
```
nocturnalastro, line 12: I think this should be binary not image.

natifridman: Done

Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+COPY --from=builder-rhel-8 /go/src/github.com/openshift/kube-compare/_output/bin/linux_arm64/kubectl-cluster_compare /usr/share/openshift/linux_arm64/kube-compare.rhel8
+COPY --from=builder-rhel-8 /go/src/github.com/openshift/kube-compare/_output/bin/linux_ppc64le/kubectl-cluster_compare /usr/share/openshift/linux_ppc64le/kube-compare.rhel8
+COPY --from=builder-rhel-8 /go/src/github.com/openshift/kube-compare/_output/bin/linux_s390x/kubectl-cluster_compare /usr/share/openshift/linux_s390x/kube-compare.rhel8
+
+COPY --from=builder-rhel-9 /go/src/github.com/openshift/kube-compare/_output/bin/linux_amd64/kubectl-cluster_compare /usr/share/openshift/linux_amd64/kube-compare.rhel9
+COPY --from=builder-rhel-9 /go/src/github.com/openshift/kube-compare/_output/bin/linux_arm64/kubectl-cluster_compare /usr/share/openshift/linux_arm64/kube-compare.rhel9
+COPY --from=builder-rhel-9 /go/src/github.com/openshift/kube-compare/_output/bin/linux_ppc64le/kubectl-cluster_compare /usr/share/openshift/linux_ppc64le/kube-compare.rhel9
+COPY --from=builder-rhel-9 /go/src/github.com/openshift/kube-compare/_output/bin/linux_s390x/kubectl-cluster_compare /usr/share/openshift/linux_s390x/kube-compare.rhel9
```
pixelsoccupied, line 33: As a final step...can we have the base image as scratch? https://hub.docker.com/_/scratch/ 

This help with any security concerns and perfect for our use case since this simply a storage for us.

natifridman: In our case, the binary depends on the arch. How can we use scratch for this?

docs/image-build.md
```diff
@@ -0,0 +1,24 @@
...
+- One option it to build the binary locally
+
+```bash
+make cross-build
+```
+
+- Another option is to extract the binary from the container
+
+```bash
+docker create --name kube-compare kube-compare:latest
```
pixelsoccupied, line 21: Could try it out? I think because we dont have any cmd specified in dockfile this will throw an error? 
```
docker create --name kube-compare kube-compare:latest ---
```

Makefile
pixelsoccupied: Sounds good!

docs/image-build.md
```diff
@@ -0,0 +1,24 @@
+
+# Kube-compare image build
+
+## Build container
+
+```bash
+make image-build
```
pixelsoccupied, line 7: now that we REAME.md updates...we can move under her https://github.com/openshift/kube-compare?tab=readme-ov-file#install ?

Dockerfile
natifridman: As I understand you should use scratch when you want to build a base image, not sure if that's our case.
Also, if I put it in the end of the dockerfile it looks like you can't access the shell.

docs/image-build.md
natifridman: I tested it with Podman and didn't get any error

docs/image-build.md
natifridman: Maybe adding a link here? https://github.com/openshift/kube-compare?tab=readme-ov-file#further-resources

docs/image-build.md
nocturnalastro: As long as its just called ContainerFile or DockerFile then it should get picked up

docs/image-build.md
pixelsoccupied: Sounds good

Dockerfile
pixelsoccupied: gotcha!

docs/image-build.md
natifridman: Done

PR #56, opened 2024-07-11
use scratch as final base image

pixelsoccupied:
Using scratch as base image to effectively remove any security concerns and use the container as simple storage for all the binaries 

https://hub.docker.com/_/scratch/ 

tested with `quay.io/npathan/dev:kc-8-2`

/cc @natifridman 
/cc @nocturnalastro

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,7 +12,7 @@ COPY . .
 RUN go work vendor
 RUN make cross-build --warn-undefined-variables
 
-FROM --platform=linux/amd64 registry.ci.openshift.org/ocp/4.17:cli
+FROM --platform=linux/amd64 registry.ci.openshift.org/ocp/4.17:cli as final-builder
 
 COPY --from=builder-rhel-9 /go/src/github.com/openshift/kube-compare/_output/bin/ /usr/share/openshift/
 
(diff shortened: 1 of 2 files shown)


natifridman:
This removes bash access in the container. I'm ok with this if you think users won't need it, should we document it somewhere?
```
Error: preparing container e2e7547e98fe6788f0ed5a5fef46d3f0e6589dfb6e09cd9aa9982e5a4370f0a9 for attach: crun: executable file `/bin/bash` not found in $PATH: No such file or directory: OCI runtime attempted to invoke a command that was not found
```

nocturnalastro:
I'm ok with removing the shell, perhaps kube-compare should be the set as the entry point?

nocturnalastro:
Was able to pull and copy the file out from `quay.io/npathan/dev:kc-8-2` although not run as I'm amd64.

nocturnalastro:
> I'm ok with removing the shell, perhaps kube-compare should be the set as the entry point?

Although which one will depend on the arch I guess :/

pixelsoccupied:
@nocturnalastro there's amd64 version in there (check the dock)

pixelsoccupied:
> Although which one will depend

Yeah this is essentially a storage now and nothing to run (except moving the binaries out to local)

nocturnalastro:
> @nocturnalastro there's amd64 version in there (check the doc)

Yeah I managed to run the binary that worked fine sorry for the confusion.

nocturnalastro:
seems nati isn't in the owners file
/approve

PR #58, opened 2024-07-22
Simplify group corrilator and kind grouping.

nocturnalastro:
- Linking the group handling variables into a single object meaning you don't have to track indexes across multiple variables, making them much easier to handle. This change also has the benefit of only keep hold of groups that have templates. 

- Removed the use age of the abstract groups.Divide function in `setLiveSearchTypes` and `setClient` for a simple map which is less lines of code and much easier to follow. 

- Removed groups sub packages as its no longer used after the above changes.


pkg/compare/correlator.go
```diff
@@ -305,3 +268,58 @@ func (c *MetricsCorrelatorDecorator) addUNMatch(cr *unstructured.Unstructured) {
...
 	c.unMatchedLock.Unlock()
 }
+
+type FieldCorrelator struct {
+	Fields    [][]string
+	hashFunc  templateHashFunc
+	templates map[string][]*ReferenceTemplate
+}
+
+func (f *FieldCorrelator) Take(templates []*ReferenceTemplate) []*ReferenceTemplate {
```
natifridman, line 278: It's not so clear from the func name what it is supposed to do

pkg/compare/compare_test.go
```diff
@@ -508,16 +507,11 @@ func getCommand(t *testing.T, test *Test, modeIndex int, tf *cmdtesting.TestFact
...
-		createGroupHashFunc([][]string{{"kind"}}),
-	)
-	resourcesByKind := lo.MapKeys(resourcesByType[0], func(value []*unstructured.Unstructured, key string) string {
-		// Converted to URL Path Format:
-		return fmt.Sprintf("/%ss", strings.ToLower(key))
-	})
+	resourcesByKind := make(map[string][]*unstructured.Unstructured)
+	for _, t := range resources {
+		key := fmt.Sprintf("/%ss", strings.ToLower(t.GetKind()))
+		resourcesByKind[key] = append(resourcesByKind[key], t)
```
natifridman, line 513: As I understand `MapKeys` also transforms to another type, does it needed here?

natifridman (commented):
Looks cleaner for me to understand

pkg/compare/compare_test.go
nocturnalastro: I think because key and the return value are both strings it doesn't do any type change.

I assume go's type system would have complained if there was an issue.

pkg/compare/correlator.go
nocturnalastro: Fair enough. My thought process was it takes the values it needs the returns the rest. Do you have a suggestion? Perhaps ClaimTemplates?

lack (approved):

PR #62, opened 2024-07-24
Fix report creation for CRs with no diffs 

AlinaSecret:
Recent changes in the compare command, changed the
 value of "diff output" for diff entries that don't contain
 any diffs from "None" to  "" (empty string). This Change
 was done only in the Json/Yaml Output, the change was
 not updated in the report creator, and resulted in
 creation of failure test cases for CRs that don't
 include diffs. This commit fixes the issue, by updating
 the report creator.

Also in separate commit:
Added a check for each test that ensures that the report creator is using  the latest format of the JSON output  of the compare command, ensuring the report creator is up to date with the command.
To support this case a testsutils package was created including test code that was located in the compare package, and now moved to the new package to allow usage in the report creator.


addon-tools/report-creator/report/create_test.go
```diff
@@ -94,27 +91,11 @@ func updateCompareOutput(t *testing.T, test Test) {
...
-			t.Fatalf("test %s failed to update golden file: %s", fileName, err)
-		}
-	}
-	expected, err := os.ReadFile(fileName)
-	if err != nil {
-		t.Fatalf("test %s failed reading .golden file: %s", fileName, err)
-	}
-	require.Equal(t, string(expected), string(value))
+	result := testutils.GetFile(t, test.getJSONPath(), testutils.RemoveInconsistentInfo(t, out.String()), update)
+	require.Equal(t, result, testutils.RemoveInconsistentInfo(t, out.String()))
```
nocturnalastro, line 95: I think these two should be here
259be8654dd520474eaeda133e5b9da4184f68c011f6ff51e40feb0515865aecR78

PR #63, opened 2024-07-24
CNF-13744: Point reference flag to config file

natifridman:
Point to config file instead of a directory

Before:
```
kubectl cluster-compare -r ./reference
```

After:
```
kubectl cluster-compare -r ./reference/metadata.yaml
```

### DIFF
--- a/pkg/compare/compare.go
+++ b/pkg/compare/compare.go
@@ -80,26 +80,25 @@ var (
 
 	compareExample = templates.Examples(`
 		# Compare a known valid reference configuration with a live cluster:
-		kubectl cluster-compare -r ./reference
+		kubectl cluster-compare -r ./reference/metadata.yaml
 		
 		# Compare a known valid reference configuration with a local set of CRs:
-		kubectl cluster-compare -r ./reference -f ./crsdir -R
+		kubectl cluster-compare -r ./reference/metadata.yaml -f ./crsdir -R
 
 		# Compare a known valid reference configuration with a live cluster and with a user config:
-		kubectl cluster-compare -r ./reference -c ./user_config
+		kubectl cluster-compare -r ./reference/metadata.yaml -c ./user_config
 
 		# Run a known valid reference configuration with a must-gather output:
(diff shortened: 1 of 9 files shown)


pkg/compare/compare.go
```diff
@@ -227,24 +226,27 @@ func (o *Options) Complete(f kcmdutil.Factory, cmd *cobra.Command, args []string
...
+	if _, err := os.Stat(o.referenceConfig); os.IsNotExist(err) && !isURL(o.referenceConfig) {
+		return fmt.Errorf(refFileNotExistsError)
 	}
 
-	if isURL(o.templatesDir) {
-		fs = HTTPFS{baseURL: o.templatesDir, httpGet: httpgetImpl}
+	referenceDir := filepath.Dir(o.referenceConfig)
+	referenceFileName := filepath.Base(o.referenceConfig)
+	if isURL(o.referenceConfig) {
+		referenceDir = strings.Replace(referenceDir, "/", "//", 1)
```
nocturnalastro, line 239: Is this becuase `filepath.Base` will turn `http://` to `http:/`?

natifridman: Yes, I think it's still nicer than URL parse

nocturnalastro: Could you add a comment to explain why its there. One of those odd things that I find comments help save people a lot of time :)

natifridman: Done

pkg/compare/testdata/ReferenceDirectoryDoesntExist/localerr.golden
```diff
@@ -1,2 +1,2 @@
-error: "Reference directory doesn't exist"
```
AlinaSecret, line 1: i think we can remove this test because now this one and "ReferenceConfigFileDoesntExist" are the same

natifridman: Done

PR #64, opened 2024-07-24
Simplify isURL

nocturnalastro:


natifridman (commented):
Another option is to use `url.ParseRequestURI(path)` and catch an error
```go
uri, err := url.ParseRequestURI(path)
if err != nil {
	return false
} else {
	return true
}
```

I guess that how it's implemented now should be ok for us

nocturnalastro:
> Another option is to use `url.ParseRequestURI(path)` and catch an error
> 
> ```go
> uri, err := url.ParseRequestURI(path)
> if err != nil {
> 	return false
> } else {
> 	return true
> }
> ```
> 
> I guess that how it's implemented now should be ok for us



> ```go
> url.ParseRequestURI
> ```

`url.ParseRequestURI` will parse normal paths without an error.

pkg/compare/httpfs.go
```diff
@@ -16,7 +16,7 @@ const defaultHttpGetAttempts = 5
 
 // isURL checks if the given path is a URL by verifying if it starts with "http://" or "https://".
 func isURL(path string) bool {
-	return strings.Index(path, "http://") == 0 || strings.Index(path, "https://") == 0
+	return strings.HasPrefix(path, "http://") || strings.HasPrefix(path, "https://")
```
AlinaSecret, line 19: @nocturnalastro i copied it from here https://github.com/kubernetes/cli-runtime/blob/master/pkg/resource/builder.go#L262

so if you want you can change it there too :laughing: 

I think the new option is better :)

natifridman (approved):
/approve

nocturnalastro:
@AlinaSecret can you approve this. It didn't work for nati for some reason perhaps we should check the owners file and make sure he's a commiter or whatever its called.

PR #71, opened 2024-07-30
Select the best match from templates in cases of multiple matches

AlinaSecret:
In cases where the reference includes two templates with the same apiVersion-kind- namespace-name 4-tuple, the command will now diff with the template that best  matches the CR. The best match behavior can be configured in the diff config. The  default behavior is to use the template with the least number of lines in the diff  with the CR. It is also possible to disable this mechanism by specifying 'None'  under the bestMatchType section in the user config.

This feature is implemented by adding an option to pass to the groupCorrelator a  function that will return the best matched template. The function will be called by  the correlator in case of multiple matches.


pkg/compare/correlator.go
```diff
@@ -228,6 +229,9 @@ func (c *GroupCorrelator) Match(object *unstructured.Unstructured) (*ReferenceTe
 		case len(templates) == 1:
 			return templates[0], nil
 		case len(templates) > 1 && multipleMatchError == nil:
+			if *c.getBestMatch != nil {
```
nocturnalastro, line 232: Would it be better to allow `getBestMatch` to return an error and have then none just always return that error?

Would allow times where is undecidable to e.g. when they both have the same number of diffs to also return multiple-match.

pkg/compare/compare.go
```diff
@@ -437,6 +456,20 @@ func extractPath(str string, pathIndex int) string {
...
+	if err != nil {
+		return nil, err
+	}
+	obj := InfoObject{
+		injectedObjFromTemplate: localRef,
+		clusterObj:              clusterCR,
+		FieldsToOmit:            temp.FieldsToOmit(o.ref.FieldsToOmit),
+		allowMerge:              temp.Config.AllowMerge,
+	}
+	return runDiff(obj, o.IOStreams, o.ShowManagedFields)
```
nocturnalastro, line 470: Does it make sense to have `runDiff` as a separate function? This is the only time its used from what I can tell.

pkg/compare/correlator.go
nocturnalastro: Infact putting it here means that you'll be doing (N+1) diffs. Where as if you just return multiple values you can just handle the logic at the higher level in the call stack and just keep the best value. 

Perhaps adding a function the to metrics gathering to add multiple match or something like that. 

Another way to get rid of the extra diff is to create a match object which can have an optional diff populated if it is you don't bother doing the diff again.

pkg/compare/compare.go
```diff
@@ -318,7 +319,25 @@ func (o *Options) setupCorrelators() error {
 		{{"apiVersion"}, {"kind"}},
 		{{"kind"}},
 	}
-	groupCorrelator, err := NewGroupCorrelator(fieldGroups, o.templates)
+	getBestMatch := func(templates []*ReferenceTemplate, CR *unstructured.Unstructured) (*ReferenceTemplate, error) {
```
lack, line 322: Is it better to have this be a named function that could be separately unit-tested as opposed to an anonymous inline function?

AlinaSecret: sounds good, moved it to a named function

pkg/compare/compare.go
AlinaSecret: missed that, removed it :)

pkg/compare/correlator.go
AlinaSecret: removed the extra diff by changing the metrics gathering

pkg/compare/parsing.go
```diff
@@ -309,3 +318,36 @@ func parseDiffConfig(filePath string) (UserConfig, error) {
...
+func getBestMatchByLines(templates []*ReferenceTemplate, cr *unstructured.Unstructured, o *Options) (*ReferenceTemplate, *bytes.Buffer, error) {
+	var bestTemp *ReferenceTemplate
+	minDiffNum := math.MaxInt
+	var minDiffOutput *bytes.Buffer
+	for _, temp := range templates {
+		diffOutput, err := diffAgainstTemplate(temp, cr, o)
+		if err != nil {
+			return nil, minDiffOutput, err
+		}
+		minDiffNum = min(bytes.Count(diffOutput.Bytes(), []byte("\n")), minDiffNum)
```
nocturnalastro, line 338: Wondering if `+` and `-` at the beginning of a line would be a better measure? If diffs dispersed throughout the manifest instead of in one clump the could throw off the count as the context the diff tool gives will inflate the number of newlines.

docs/user-guide.md
```diff
@@ -129,6 +129,44 @@ After running `kubectl cluster-compare` tool, the reference configuration CRs wi
...
+
+In cases where the reference includes two templates with the same apiVersion-kind-namespace-name 4-tuple, by the correlation
+logic that was explained in previous section the command will diff with the template that is best matched to the CR.
+The best match behaviour can be configured in the diff config. The default behaviour is to use the template with the least
+number of lines in the diff with the CR (Named 'NumLines'). Is also possible to disable this mechanism by specifying 'None'
+under the bestMatchType section in the user config. The only supported values are: NumLines and None. More mechanisms can
+be added by request.
+
+```yaml
+bestMatchType: "None"
```
nocturnalastro, line 163: Sorry I missed this last time I looked. This seems more like a choice that should be made my the reference designer than the end user.

AlinaSecret: i think its more of a user decision because it depends on the diff tool used by the user that is passed in the env var.

nocturnalastro: Perhaps how to calculate the number of diffs is diff to dependent.

If there are templates which need this is reference implementation decision. 

Perhaps using something like generating a merge patch and counting the number of leaf nodes in the yaml would be more tool agnostic metric.

pkg/compare/parsing.go
nocturnalastro: Perhaps using something like generating a merge patch and counting the number of leaf nodes in the yaml would be more tool agnostic metric.

AlinaSecret:
@nocturnalastro 
Removed the option to specify the diff count mechanism in the user config,
because it may not be needed in the future once a more diff tool agnostic diff count method will be introduced (will open a story for that)

pkg/compare/testdata/WhenUsingDiffAllFlag-AllUnmatchedResourcesAppearInSummary/localout.golden
```diff
@@ -1,21 +1,62 @@
...
+     k8s-app: kindnet
+     tier: node
++  name: Test2
+   namespace: SomeNS
+
+**********************************
+
 Summary
-CRs with diffs: 0/0
+CRs with diffs: 3/1
```
AlinaSecret, line 47: just  a comment, leaving the test as is because its not relevant to this PR 
but i think this looks funny and we should change it to mkae more sense in case there are multiple CRs from the same template kind

nocturnalastro: Why is that happening if your only matching when you have gotten the match?

lack: I intuit that `3/1` means we found 3 CRs that matched a single template?

If that's what this means, I agree it looks weird and maybe is an indication that the message format and/or not what a user would expect.

So let's open that other story, link it here for reference, and go ahead with this change.

PR #72, opened 2024-07-30
CNF-13833: Initialize go vendoring for this repository

lack:
Go vendoring protects us against unexpected failures of external sites and allows for fully offline builds of this tool.


natifridman:
Maybe after this PR merged we should add a test like we have for go mod tidy? https://github.com/openshift/release/blob/d1382f7228a74c983a9a4955c2e6c524ba76c620/ci-operator/config/openshift/kube-compare/openshift-kube-compare-main.yaml#L90

lack:
Interesting...  The tool is now flagging potential vulnerabilities in the go modules, which it apparently wasn't doing before.

I don't know a lot about the scanner... Is it possible to only issue non-fatal warnings in a particular path?  Or failing that, configure it to skip over vendor/**?

natifridman:
@lack It's possible to ignore an issue in Snyk UI. The link is available in the security Prow job

nocturnalastro:
/lgtm

You guys have more of an idea of snyck than I do what do you think?
/cc @natifridman @pixelsoccupied

PR #80, opened 2024-08-08
Add example of a catch all template

nocturnalastro:
Example of how you can use less specific templates to act as a catch all


pkg/compare/testdata/MachineConfigsCatchAll/reference/other_mcs.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: machineconfiguration.openshift.io/v1
+kind: MachineConfig
+metadata:
+    name: {{ .metadata.name }}
```
natifridman, line 4: Maybe we can also add something to catch the name? not sure about the syntax, but something like this.
```
{{ if .metadata.name | contains "container-mount-"}}
```

nocturnalastro: The name is meant to be a catch all here, we have templates for the container-mounts 
```yaml
- path: 01-container-mount-ns-and-kubelet-conf-master.yaml
- path: 01-container-mount-ns-and-kubelet-conf-worker.yaml
```

which we want to be specific becuase it means they get caught by the corrilator before the catch all template.

Think of it like a switch statemnet
```go
switch templateName {
  case "container-mount-namespace-and-kubelet-conf-master":
     return "01-container-mount-ns-and-kubelet-conf-master.yaml"
  case "container-mount-namespace-and-kubelet-conf-worker":
     return "01-container-mount-ns-and-kubelet-conf-worker.yaml"
  default:
     return "other_mcs.yaml"
}
```

natifridman: Thanks, so the idea is to catch all machine configs and not just for the container-mounts?

nocturnalastro: It shouldn't catch the `container-mounts` as they should be caught by the specific cases.

lack (approved):

PR #81, opened 2024-08-09
Fix parse template metadata nil

nocturnalastro:


AlinaSecret:
/approve
great changes :) 
thanks @nocturnalastro

PR #88, opened 2024-08-22
Onboarding new component to OCP

lgarciaaco:
Ref ticket: https://issues.redhat.com/browse/ART-10503

 - Use openshift-enterprise-cli as base
 - Remove `go work vendor` before build step
 - Ensure -mod=vendor when building


lack (approved):
Yes, since we're vendoring in the repo now, there's no need for the Dockerfile to re-vendor.

PR #93, opened 2024-08-29
CNF-11871: Allow user overrides

nocturnalastro:


lack:
I think the change looks great, but probably necessitates some doc changes too?

pkg/compare/testdata/UserOverride/localout.golden
```diff
@@ -0,0 +1,46 @@
...
+Diff Output: None
+Patched with testdata/UserOverride/localnewOverridesout.golden
+Patch Reasons: <None given>
+
+**********************************
+
+Summary
+CRs with diffs: 1/2
+No CRs are missing from the cluster
+No CRs are unmatched to reference CRs
```
lack, line 46: Maybe the Summary should also include a line like `"%d CRs are locally patched"` or something like that?

pkg/compare/testdata/UserOverride/localout.golden
```diff
@@ -0,0 +1,46 @@
...
++  finalizers:
++  - kubernetes
+
+**********************************
+
+Cluster CR: v1_Namespace_openshift-storage
+Reference File: namespace.yaml
+Diff Output: None
+Patched with testdata/UserOverride/localnewOverridesout.golden
+Patch Reasons: <None given>
```
lack, line 39: Should we maybe fail in case of "None given" to force some documentation of the patches?

lack (changes requested):
Generally really good, but we need doc updates too before we merge this please!

pkg/compare/testdata/UserOverride/localout.golden
nocturnalastro: Do you mean fail to generate them or fail to report on a diff?

pkg/compare/testdata/UserOverride/localout.golden
nocturnalastro: I like that idea:)

pkg/compare/testdata/UserOverride/localout.golden
nocturnalastro: We could refuse to add the patch if no reason is given and then put an error/warning log to explain

nocturnalastro:
@lack I added some docs in explaining something I realised I'd missed at test so I added that and fixed the bug it showed :).

pkg/compare/testdata/UserOverride/localout.golden
nocturnalastro: Done

pkg/compare/testdata/UserOverride/localout.golden
lack: I still think forcing "some string" is better than allowing nothing.  We can't stop people from lying about the reason, but we can at least encourage some kind of good behavior.

lack (approved):
Looks great!

PR #102, opened 2024-09-12
CNF-14198: Add version to metadata but default to v1 if missing

nocturnalastro:


natifridman:
Maybe we can separate the metadata structs into files for readability? so we can have something like `metadataV1.go` and `metadataV2.go`

lack (approved):
Awesome!

PR #108, opened 2024-09-17
CNF-13832 Metadata apiv2

nocturnalastro:


lack:
I noticed two things generally, that we should be careful about:
- The first commit has some go module bumps plus some code changes.  Let's try to keep those in separate commits (if not separate PRs when possible)
- Because we have a go module bump, you'll need to run `go work vendor` to update the vendor/ copies of our modules, and then don't forget to `git add vendor/` to get all the changes in the same commit as the go.mod changes.

nocturnalastro:
> I noticed two things generally, that we should be careful about:
> 
>     * The first commit has some go module bumps plus some code changes.  Let's try to keep those in separate commits (if not separate PRs when possible)
> 
>     * Because we have a go module bump, you'll need to run `go work vendor` to update the vendor/ copies of our modules, and then don't forget to `git add vendor/` to get all the changes in the same commit as the go.mod changes.

I moved them to a separate PR, I was checked the sync and vendor before I started this must have forgotten to clear them.

docs/reference-config-guide-v2.md
```diff
@@ -0,0 +1,329 @@
...
+  - name: ExamplePart2
+    ---- here goes Part2 configuration ----
+```
+
+"components" array includes objects with the following keys:
+
+- a pair at key "name" with a value of a string identifying a component required for the part
+- one of the following:
+  - `allOf`: Will cause a validation error if not all templates are matched for required templates.
+  - `allorNoneOf`: Will cause a validation error when one or more but not all templates are matched for optional blocks of templates.
```
lack, line 60:
```suggestion
  - `allOrNoneOf`: Will cause a validation error when some but not all of the templates are matched. Use for sets of templates where the entire set is optional, but if at least one is present then all the others must also be present.
```

docs/reference-config-guide-v2.md
```diff
@@ -0,0 +1,329 @@
...
+        ---- here goes ExampleComponent1 configuration ----
+  - name: ExamplePart2
+    ---- here goes Part2 configuration ----
+```
+
+"components" array includes objects with the following keys:
+
+- a pair at key "name" with a value of a string identifying a component required for the part
+- one of the following:
+  - `allOf`: Will cause a validation error if not all templates are matched for required templates.
```
lack, line 59:
```suggestion
  - `allOf`: Will cause a validation error if not all the templates are matched. Use for sets of required templates.
```

docs/reference-config-guide-v2.md
```diff
@@ -0,0 +1,329 @@
...
+  - `allorNoneOf`: Will cause a validation error when one or more but not all templates are matched for optional blocks of templates.
+  - `anyOf`: Will never cause a vaildation error, for completely optional templates
```
lack, line 61:
```suggestion
  - `anyOf`: Will only cause validation errors for the content of matched templates. Use for completely optional templates.
```

docs/reference-config-guide-v2.md
```diff
@@ -0,0 +1,329 @@
...
+  - `noneOf`: Will cause a vaildation error if any of the templates are matched useful to exclude CRs
```
lack, line 62:
```suggestion
  - `noneOf`: Will cause a validation error if any of the templates are matched. Use for excluding CRs.
```

docs/reference-config-guide-v2.md
```diff
@@ -0,0 +1,329 @@
...
+  - `oneOf`: Will cause a vaildation error when None or more than one templates is matched, for when you want to match exactly one template in the block
```
lack, line 63:
```suggestion
  - `oneOf`: Will cause a validation error when none or more than one of the templates is matched. For requiring exactly one template in the set.
```

docs/reference-config-guide-v2.md
```diff
@@ -0,0 +1,329 @@
...
+  - `anyOneOf`: Will throw a validation error if more than one is matched, for when you optionally want only one of the templates the block
```
lack, line 64:
```suggestion
  - `anyOneOf`: Will throw a validation error if more than one template is matched. Use for optionally allowing only one of the templates in the set.
```

docs/reference-config-guide-v2.md
```diff
@@ -0,0 +1,329 @@
...
+
+```yaml
+# requiredTemplates contains all needed reference CRs to comply with ExampleComponent1
```
lack, line 67: I think this comment is wrong now?

lack (changes requested):
Doc only review at first; I think we should use `allOrNoneOf` instead of `allorNoneOf` or the camelCasePolice may start knocking down our doors.

lack (approved):

PR #109, opened 2024-09-18
Sync and vendor the workspace

nocturnalastro:


lack:
I don't see any changes under `vendor/**` and I would have expected there to be some...

lack:
Also it looks like the point of this commit is (and therefore the commit log message should be) "version-bump to a hand full of go modules"; the sync is "how" not "what" ;)

nocturnalastro:
I think its because all I ran was `go work sync` instead of actually adding a new version into the repo. 

You can try youself on main
```
go work sync
go work vendor
```

nocturnalastro:
From the go workspaces docs
```
go work sync syncs dependencies from the workspace’s build list into each of the workspace modules.
```
I think the point is you don't get differences in the versions within the workspace.

go.mod
```diff
@@ -3,27 +3,27 @@ module github.com/openshift/kube-compare
 go 1.22.0
 
 require (
-	github.com/BurntSushi/toml v0.3.1
+	github.com/BurntSushi/toml v1.3.2
```
lack, line 6: I find it troubling that our root-level go.mod has a version bump but there's no accompanying change to `vendor/github.com/BurntSushi/toml` - I would expect we'd need both?

lack: Ah, I see - One of the submodluse must have had the newer version, so the vendor/ area already had it...

PR #111, opened 2024-09-19
CNF-14097: Count leaves of the patch to find smallest diff

nocturnalastro:
In the case where a patch can not be made for one of the templates it will fallback to the old method of counting the number of lines

### DIFF
--- a/pkg/compare/compare.go
+++ b/pkg/compare/compare.go
@@ -8,7 +8,6 @@ import (
 	"errors"
 	"fmt"
 	"io/fs"
-	"math"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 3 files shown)


lack:
I don't see a difference in any tests; do we have a demonstration of what is different because of this change?

nocturnalastro:
I had to change one of the sets of tests because it seemed to be relying on the order of processing.

https://github.com/openshift/kube-compare/pull/111/files#diff-b2203574e46de327d642a8e5d08f6aeaf322ea4cb69bd56482d061751a256ed7R7

lack (approved):

PR #112, opened 2024-09-25
Add krew-bump.sh helper script for updating the krew index as part of release

lack:


natifridman (commented):
Maybe we can use Krew auto update? https://krew.sigs.k8s.io/docs/developer-guide/release/automating-updates/

lack:
@natifridman : I looked into that automation, but it only seems to be designed to work with github actions and I'm not sure that jives with the openshift org prow integration :(

natifridman (approved):
/lgtm

PR #125, opened 2024-10-29
Add regex inline diff function

AlinaSecret:
This update adds the option to include common inline diff
 functions in the tool.  Diff functions may simplify complex 
diff logic, which would be hard to write and read using  regular Go templating and include a more tidy usage.

A new regex inline diff function allows checking fields in cluster resources against a regex pattern. If a  field matches the regex, no diffs are shown; if it doesn’t  match, a diff will highlight the difference. To use this,  enable `regexInline` for the field in `metadata.yaml`
 and add the regex in the template.

([CNF-14934](https://issues.redhat.com//browse/CNF-14934))

### DIFF
--- a/docs/reference-config-guide-v2.md
+++ b/docs/reference-config-guide-v2.md
@@ -344,3 +344,69 @@ metadata:
 ```
 
 you use would use `metadata.annotations."workload.openshift.io/allowed"`.
+
+### PerField Configuration
+
+#### Inline Diff Funcs
+
+This Version of the tool contains builtin Inline diff functions. Inline diff functions are functions that run on a specific
+field in the templates. The functions contain common advanced diff functions that may be also implemented with pure go templating
+but if they were implemented with go templating would result in unclear diff logic that is hard to maintain.
+The functions goal is to enable a more clear and readable usage for common custom templating functions and allow more complex
+functionalities.
+
+To specify an inline function for a specific field yse the `perField` section in the metadata.yaml file as in this example:
+
+```yaml
+apiVersion: v2
+parts:
+- name: ExamplePart
+  components:
+  - name: Example
+    allOf:
+    - path: cm.yaml
+      config:
+        perField:
+        - pathToKey: spec.bigTextBlock # Field in template to run the inline diff function in pathToKey syntax
+          inlineDiffFunc: regex # Inline function 
+```
+
+Supported inline diff functions:
+
+##### Regex Inline Diff Function
+
+The regex Inline diff function allows validating fields in CRs based on a regex. When using the function the command will
+show no diffs in case the cluster CR will match the regex, if it does not match the regex a diff will be shown between the
+cluster CR and the regex expression.
+To use the regex inline diff function you need to enable the regexInline function for the specific field and template in
+the metadata.yaml and also specify the regex inside the template.
+For example:
+
+For a template named cm.yaml where spec.bigTextBlock should be validated by regex:
+
+```yaml
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  namespace: kubernetes-dashboard
+spec:
+  bigTextBlock: |-
+    This is a big text block with some static content, like this line.
+    It also has a place where (?<username>[a-z0-9]+) would put in their own name. (?<username>[a-z0-9]+) would put in their own name.
+```
+
+the metadata.yaml should contain:
+
+```yaml
+apiVersion: v2
+parts:
+- name: ExamplePart
+  components:
+  - name: Example
+    allOf:
+    - path: cm.yaml
+      config:
+        perField:
+        - pathToKey: spec.bigTextBlock 
+          inlineDiffFunc: regex  
+```
--- a/pkg/compare/compare.go
+++ b/pkg/compare/compare.go
@@ -614,6 +614,7 @@ func diffAgainstTemplate(temp ReferenceTemplate, clusterCR *unstructured.Unstruc
 		FieldsToOmit:            temp.GetFieldsToOmit(o.ref.GetFieldsToOmit()),
 		allowMerge:              temp.GetConfig().GetAllowMerge(),
 		userOverrides:           userOverrides,
+		templateFieldConf:       temp.GetConfig().GetInlineDiffFuncs(),
 	}
 
 	differ, err := diff.NewDiffer("MERGED", "LIVE")
@@ -672,7 +673,7 @@ func (o *Options) Run() error {
 			klog.Warningf(skipInvalidResources, extractPath(err.Error(), 2), err.Error()[strings.LastIndex(err.Error(), ":"):])
 			return true
 		}
-		return containOnly(err, []error{UnknownMatch{}, MergeError{}})
+		return containOnly(err, []error{UnknownMatch{}, MergeError{}, InlineDiffError{}})
 	})
 
 	err := r.Visit(func(info *resource.Info, _ error) error { // ignoring previous errors
@@ -758,6 +759,7 @@ type InfoObject struct {
 	FieldsToOmit            []*ManifestPathV1
 	allowMerge              bool
 	userOverrides           []*UserOverride
+	templateFieldConf       map[string]inlineDiffType
 }
 
 // Live Returns the cluster version of the object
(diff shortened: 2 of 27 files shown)


pkg/compare/referenceV2.go
```diff
@@ -214,6 +217,85 @@ func (entry *FieldsToOmitV2Entry) process(previousKeys []string, toOmit *FieldsT
...
+}
+
+type inlineDiffType string
+
+const (
+	regex inlineDiffType = "regex"
+)
+
+var InlineDiffFuncs = map[inlineDiffType]func(string, string) string{regex: regexDiffFunc}
+var InlineDiffFuncsValidators = map[inlineDiffType]func(string) error{regex: regexDiffFuncValidator}
```
nocturnalastro, line 278: I think these should actually be a single struct with func(probably diff or something) and validator as fields on it as they are required to come in pairs.

docs/reference-config-guide-v2.md
```diff
@@ -344,3 +344,69 @@ metadata:
...
+
+```yaml
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  namespace: kubernetes-dashboard
+spec:
+  bigTextBlock: |-
+    This is a big text block with some static content, like this line.
+    It also has a place where (?<username>[a-z0-9]+) would put in their own name. (?<username>[a-z0-9]+) would put in their own name.
```
nocturnalastro, line 395: Is example seems like we're doing something with the named capture groups which we're not. 

I think it would be a good idea to extend to that perhaps and thing that is named the same should be checked to be equal or it throws an error.

lack: I like this idea of enforcing that all named capture groups are equivalent!  Maybe in a second PR for this story?

pkg/compare/compare.go
```diff
@@ -792,11 +794,33 @@ func (obj InfoObject) Merged() (runtime.Object, error) {
...
 
 	omitFields(obj.injectedObjFromTemplate.Object, obj.FieldsToOmit)
 	return obj.injectedObjFromTemplate, err
 }
 
+func (obj InfoObject) runInlineDiffFuncs() {
+	for pathToKey, inlineDiffFunc := range obj.templateFieldConf {
+		listedPath, err := pathToList(pathToKey)
+		if err != nil {
+			return
```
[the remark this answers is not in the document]
lack: Or if we do `return`, that would mean it's a syntax error in the metadata or the template, right?  Perhaps this function should return `error` so we can output a relevant message to the user?

pkg/compare/testdata/ReferenceV2InlineRegex/localwithDiffout.golden
```diff
@@ -0,0 +1,22 @@
...
+Reference File: cm-regex-with-diff.yaml
+Diff Output: diff -u -N TEMP/v1_configmap_kubernetes-dashboard_kubernetes-dashboard-settings TEMP/v1_configmap_kubernetes-dashboard_kubernetes-dashboard-settings
+--- TEMP/v1_configmap_kubernetes-dashboard_kubernetes-dashboard-settings	DATE
++++ TEMP/v1_configmap_kubernetes-dashboard_kubernetes-dashboard-settings	DATE
+@@ -8,4 +8,4 @@
+ spec:
+   bigTextBlock: |-
+     This is a big text block with some static content, like this line.
+-    It also has a place where (?<username>[0-9]+) would put in their own name. (?<username>[0-9]+) would put in their own name.
++    It also has a place where exampleuser would put in their own name. exampleuser would put in their own name.
```
lack, line 13: Can you help me understand this test case?  What are we testing here?

pkg/compare/compare.go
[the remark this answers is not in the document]
nocturnalastro: I think continue and compiling a list errors would be a good idea.

docs/reference-config-guide-v2.md
nocturnalastro: I think you could do it with a loop through the capture groups [here](https://github.com/openshift/kube-compare/pull/125/files#diff-7776a9c2913a158bf06991b034112f09595963593730e030e94249a740df4358R285)

pkg/compare/testdata/ReferenceV2InlineRegex/localwithDiffout.golden
lack, line 13: Is there a way we could make it more clear that the diff is because of the regex mismatch and not some other difference?  For example, could we expand the template side to have something like `<<exampleuser does not match (?<username>[0-9+)>>`?

pkg/compare/testdata/ReferenceV2InlineRegex/resources/cm.yaml
lack, line 1: I think we need one more test case, one where the 1st line is wrong but the 2nd line matches properly.  Something like
```yaml
kind: ConfigMap
apiVersion: v1
metadata:
  labels:
    k8s-app: kubernetes-dashboard
  name: kubernetes-dashboard-settings
  namespace: kubernetes-dashboard
spec:
  bigTextBlock: |-
    This is a big text block with some content that doesn't match.
    It also has a place where exampleuser would put in their own name. exampleuser would put in their own name.
```

docs/reference-config-guide-v2.md
lack: Detecting the error would be not too bad.  Once we detect a group mismatch, how do we munge the template to indicate the error?

pkg/compare/testdata/ReferenceV2InlineRegex/localwithDiffout.golden
nocturnalastro: I think that should be in a future story.

docs/reference-config-guide-v2.md
nocturnalastro: We would probably need to use the `regexp/syntax` but I think should be another story. To improve the way we report it in general

nocturnalastro: Perhaps we could add a reason below the diff in the output saving us having to mangle textblob?

pkg/compare/testdata/ReferenceV2InlineRegex/localwithDiffout.golden
nocturnalastro: regex match won't tell you where it failed so you would probably have to enumerate the sub expressions until something failed.

lack: Agreed.  Future PR.

pkg/compare/testdata/ReferenceV2InlineRegex/resources/cm.yaml
AlinaSecret: added the testcase, though i think for now it does not have much value. but in the future, once there is better description for the reason of the diff, then the test case will already exist

pkg/compare/compare.go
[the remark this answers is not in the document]
AlinaSecret: done, though some of these errors will never occur (or at least should) because they are detected when parsing the templates in the validation function

docs/reference-config-guide-v2.md
AlinaSecret: i think instead of bellow the diff, you can change the output of the inline regex function to set the field to be the regex with a message according to the issue. that way it will have much more context in the diff because it will appear in the place of the problematic field. instead a list of errors after the diff

nocturnalastro: Yeah it will require introspection of the regex.

pkg/compare/testdata/ReferenceV2InlineRegex/localWithDiffInFirstLineout.golden
```diff
@@ -0,0 +1,24 @@
...
+--- TEMP/v1_configmap_kubernetes-dashboard_kubernetes-dashboard-settings	DATE
++++ TEMP/v1_configmap_kubernetes-dashboard_kubernetes-dashboard-settings	DATE
+@@ -7,5 +7,5 @@
+   namespace: kubernetes-dashboard
+ spec:
+   bigTextBlock: |-
+-    This is a big text block with some static content, like this line
+-    It also has a place where (?<username>[a-z0-9]+) would put in their own name. (?<username>[a-z0-9]+) would put in their own name.
++    This is a big text block with some static content, like this line.
++    It also has a place where exampleuser would put in their own name. exampleuser would put in their own name.
```
lack, line 15: I can see why the 1st line is different; the missing period at the end.  But why is this 2nd line showing up as different?

AlinaSecret: that is why i said the test probably does not have enough value. because in this version the function only check the if the expression matches the regex   and if it does not the value that will be in template is the regex if it does match then it will be set to the value in the cluster cr resulting in no diffs.
for future work we can upgrade the mechanism to return no the regex but divvide it in to lines and run the regex separately per each line and then return a blob that will partially contain the regex and partially the clusters data.
this feature needs a little bit of thinking because what happens if a line is missing and other scenarios

lack: Ah, I see!  Perhaps I was expecting we would do a line-by-line match, but I understand why we're not yet.

lack (approved):

lack:
I think this is a great start!  We may enhance it to do line-by line diffs or something clever with the capture groups, but that's for another day.

PR #129, opened 2024-11-02
CNF-15499: Allow references to supply per-template description text

lack:

### DIFF
--- a/docs/reference-config-guide-v2.md
+++ b/docs/reference-config-guide-v2.md
@@ -88,6 +88,30 @@ components:
       - path: OptionalExclusiveTemplate2.yaml
 ```
 
+### Reference Descriptions
+
+In order to make detected differences more actionable, each template may
+include a description section which is displayed when a difference is detected,
+or if a required template is absent.
+
+The description section is a free-form multiline text field which can contain
+instructions or a URL referencing extra documentation about the template in
+question, such as why it is required or which fields are optional.
+
+```yaml
+components:
+  - name: ExampleComponent1
+    allOf:
+      - path: RequiredTemplate1.yaml
+        description: |-
+          Required for Important Reasons.
+          See https://github.com/org/repo/doc.md for details.
+      - path: RequiredTemplate2.yaml
+        description: |-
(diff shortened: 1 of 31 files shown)


docs/reference-config-guide-v2.md
```diff
@@ -88,6 +88,32 @@ components:
...
+which can contain instructions or a URL referencing extra
+documentation about the template in question, such as why it
+is required or which fields are optional.
+
+```yaml
+components:
+  - name: ExampleComponent1
+    allOf:
+      - path: RequiredTemplate1.yaml
+        helpText: |-
```
natifridman, line 108: Should we call it `description` instead?

pkg/compare/referenceV1.go
```diff
@@ -204,6 +214,7 @@ func (config ReferenceTemplateConfigV1) GetFieldsToOmitRefs() []string {
 type ReferenceTemplateV1 struct {
 	*template.Template `json:"-"`
 	Path               string                    `json:"path"`
+	Description        string                    `json:"description,omitempty"`
```
nocturnalastro, line 217: I think this should be on ReferenceTemplateV2. And the V1 template should just return an empty string.

lack: I think it's "safe" to add new optional fields to older APIs though, and V1 could benefit from this as well at no extra cost.

nocturnalastro: Yeah shouldn't be an issue I suppose. 
Could you add a test for it then just to make sure we don't break `getMissingCRs` for anyone who decided to go with V1 and use the new field.

lack: Suddenly "no extra cost" became "adding another test", so I'm now less convinced ;)

lack:
/override ci/prow/unit

The intermittent failure here is not related to this change.

pkg/compare/referenceV1.go
lack: Test added.

natifridman (approved):
/lgtm

PR #130, opened 2024-11-05
Handle capture groups in regex blob

nocturnalastro:


lack:
I think this was replaced by #137

PR #132, opened 2024-11-07
Escape values if refernce filenames contain '.'

lack:
Also adds unit tests for all currenty-supported filename escapes.


natifridman (approved):
/lgtm

PR #137, opened 2024-11-18
compare of named capture group values in inline regex

nocturnalastro:


pkg/compare/regexInlineDiff_test.go
lack, line 1: Yes!  We need more actual unit tests!

PR #138, opened 2024-11-19
unstructured nested list traversal

lack:
- **Add stacktrace to test panic output**
- **Replace unstructured.NestedString with a copy that allows list traversal**
- **Perform field validation just-in-time for diff**
- **Excercise list traversal pathToKey for capturegroups testing**


pkg/compare/compare_test.go
```diff
@@ -606,7 +607,7 @@ func TestCompareRun(t *testing.T) {
 				defer func() {
 					r := recover()
 					if s, ok := r.(string); r != nil && (!ok || s != ExpectedPanic) {
-						t.Fatalf("test paniced: %v", r)
+						t.Fatalf("test paniced: %v\n%s", r, string(debug.Stack()))
```
nocturnalastro, line 614: <3

PR #141, opened 2024-11-20
OCPBUGS-43700: Fix reporting of diff when using external diff 'diff -y'

nocturnalastro:
Beacuse 'diff -y' returns the entirity of both manifest even when no diffs are present, had to switch to using attempting to use the leaf count and not the diff output to determing if a diff exists in the output


lack (commented):
As an alternative, could we rely on the return code of the diff execution?

lack:
Should we add a compare_test scenario for this bad boy?

PR #143, opened 2024-11-22
Revert "Introduce strict-match correlation flag"

lack:
This reverts commit 02692f3b8ab0e7777da66b196f2a2f8f651e8f05.

Turns out on live systems this change is VERY BAD.  For example, it tries to match all MachineConfig objects to the few we have in the reference, resulting in thousands of false negatives.  We'll have to rethink the approach for this issue.


imiller0 (commented):
/lgtm

PR #167, opened 2025-01-29
Remove go workspace from source control

lack:
This consolidates all addon-tools under the same go module control for
easier dependency management.

It will also allow dependabot to be more useful :)


natifridman (commented):
This will make life easier 😃 
/lgtm

PR #172, opened 2025-02-12
CNF-14495: Add ability to retrieve references directly from container

adhil0:
In the future, it will be preferable for us to deliver references to users via a container. Give users the ability to refer to a
metadata.yaml in an image directly from the kube-compare command-line. 

Use `exec.Command()` to run podman and docker operations rather than official go bindings because quick experimentation showed that it would increase the binary size from ~90 MB to ~120 MB. 

Still need to address:

- [x] podman/docker login to registry; Don't need to login, just need to make sure error messages are descriptive
- [x] cleaning up files created in tmp directory
- [x] add more tests


pkg/compare/container.go
```diff
@@ -0,0 +1,139 @@
...
+		return c == ':'
+	}
+	sections := strings.FieldsFunc(path, f)
+
+	if len(sections) == 3 {
+		image := sections[0] + ":" + sections[1]
+		referencePath := sections[2]
+		return image, referencePath, nil
+	}
+	return "", "", fmt.Errorf("Incorrect path passed into -r, it should follow this format: container://<IMAGE>:<TAG>:/path_to_metadata.yaml")
```
lack, line 39:
```suggestion
	return "", "", fmt.Errorf("Incorrect container path passed to -r, it must follow this format: container://<IMAGE>:<TAG>:/path_to_metadata.yaml")
```

README.md
```diff
@@ -56,6 +56,12 @@ Run a known valid reference configuration with an `oc must-gather` output:
 kubectl cluster-compare -r ./reference/metadata.yaml -f "must-gather*/*/cluster-scoped-resources","must-gather*/*/namespaces" -R
 ```
 
+Retrieve a known valid reference configuration from a container and compare with a local set of CRs
```
lack, line 59:
```suggestion
Extract a known valid reference configuration from a container and compare with a local set of CRs:
```

pkg/compare/compare.go
```diff
@@ -87,6 +87,9 @@ var (
 
 		# Run a known valid reference configuration with a must-gather output:
 		kubectl cluster-compare -r ./reference/metadata.yaml -f "must-gather*/*/cluster-scoped-resources","must-gather*/*/namespaces" -R
+
+		# Retrieve a known valid reference configuration from a container and compare with a local set of CRs:
```
lack, line 91:
```suggestion
		# Extract a known valid reference configuration from a container image and compare with a local set of CRs:
```

pkg/compare/container.go
```diff
@@ -0,0 +1,140 @@
...
+}
+
+// isContainer reports whether the given path is a reference to a file in a container by verifying if it starts with "container://".
+func isContainer(path string) bool {
+	return strings.HasPrefix(path, "container://")
+}
+
+// parsePath returns the image and referencePath (path to the directory for metadata.yaml), given a path
+// of the form container://<IMAGE>:<TAG>:/path_to_metadata.yaml
+func parsePath(path string) (string, string, error) {
```
lack, line 24: Returning a tuple can sometimes be confusing, because you have to remember which string is which.  Sometimes it's valuable to create a struct with named arguments instead:

```suggestion
type parsedPath struct {
    image string
    path  string
}

func parsePath(path string) (parsedPath, error) {
```

pkg/compare/container.go
```diff
@@ -0,0 +1,140 @@
...
+
+	if len(sections) == 3 {
+		image := sections[0] + ":" + sections[1]
+		referencePath := sections[2]
+		return image, referencePath, nil
+	}
+	return "", "", fmt.Errorf("incorrect path passed to -r, it must follow this format: container://<IMAGE>:<TAG>:/path/to/metadata.yaml")
+}
+
+// Use var's so that we can mock functions in tests.
```
lack, line 43: Spectacular!

pkg/compare/container.go
```diff
@@ -0,0 +1,140 @@
...
+	engine.tempDir = filepath.Join(dname, filepath.Base(pathToMetadata))
+	return nil
+}
+
+// cleanup stops and removes the container used to extract the reference configs.
+func (engine *engine) cleanup() error {
+	out, err := engine.runEngineCommand("stop", engine.containerID)
+	if err != nil {
+		// Print errors as warnings rather than returning, since stopping and removing the container is not vital.
+		fmt.Printf("Warning: Could not stop container: %s", out)
```
lack, line 105: Change these fmt.Printf lines to klog.Warningf for more consistent output:
```suggestion
		klog.Warningf("Could not stop container: %s", out)
```

pkg/compare/container.go
```diff
@@ -0,0 +1,140 @@
...
+	}
+	out, err = engine.runEngineCommand("rm", engine.containerID)
+	if err != nil {
+		fmt.Printf("Warning: Could not remove container: %s", out)
```
lack, line 109:
```suggestion
		klog.Warningf("Could not remove container: %s", out)
```

pkg/compare/container.go
```diff
@@ -0,0 +1,140 @@
...
+	}
+	return nil
+}
+
+// getReferencesFromContainer uses a path to an image and a metadata.yaml within that image, and extracts the reference configs
+// to a local temporary directory. Returns the path to this directory.
+func getReferencesFromContainer(path string, tempContainerRefDir string) (string, error) {
```
lack, line 116: Minor nit, but some code linters may complain about this - If you have 2 args of the same type in a row, you can combine them like this:
```suggestion
func getReferencesFromContainer(path, tempContainerRefDir string) (string, error) {
```

pkg/compare/container_test.go
```diff
@@ -0,0 +1,326 @@
...
+
+func TestIsContainer(t *testing.T) {
+	tests := []struct {
+		input    string
+		expected bool
+	}{
+		{"container://my-image:latest:/metadata.yaml", true},
+		{"file://local/path/to/file.yaml", false},
+		{"http://example.com/data.yaml", false},
+		{"randomstringwithoutprefix", false},
```
lack, line 21: We could consider adding other potential edge cases such as:
- `"container:/only-one-slash:tag:/metadata.yaml"`
- `"container/with/no/colon/metadata.yaml"`
- `"container://name:tag"`
- `"container://name"`
- `"container://"`
- `"container`
- `""`

pkg/compare/container_test.go
```diff
@@ -0,0 +1,326 @@
...
+		expectedImage string
+		expectedPath  string
+		expectError   bool
+	}{
+		{"container://my-image:latest:/metadata.yaml", "my-image:latest", "/metadata.yaml", false},
+		{"container://repo/image:tag:/data.yaml", "repo/image:tag", "/data.yaml", false},
+		{"container://repo/image:::tag::::/data.yaml", "repo/image:tag", "/data.yaml", false},
+		{"container://wrongformat", "", "", true},
+		{":missingprefix:/file.yaml", "", "", true},
+		{"container://only:two", "", "", true},
```
lack, line 44: As above, try to think like a user who has a 2-year-old helping them type commands in ways that would potentially confuse the parser:
- `"container://"`
- `"container:/"`
- `"container:"`
- `"container"`
- `""`

pkg/compare/container_test.go
```diff
@@ -0,0 +1,326 @@
...
+		lookPath = exec.LookPath
+	}()
+
+	tests := []struct {
+		available    string
+		expectedName string
+		expectedSudo bool
+		expectError  bool
+	}{
+		{"podman", "podman", false, false},     // Podman is available
```
lack, line 153: One more case I thought of - What if both `podman` and `docker` are available?

pkg/compare/container_test.go
```diff
@@ -0,0 +1,326 @@
...
+
+	tests := []struct {
+		engine         engine
+		pathToMetadata string
+		dname          string
+		expectError    bool
+	}{
+		{engine{"podman", false, "container123", ""}, "/etc/configs", "/tmp/extracted", false},
+		{engine{"docker", true, "container456", ""}, "/var/lib/data", "/tmp/extracted", false},
+		{engine{"docker", true, "container456", ""}, "/invalid_path", "/tmp/extracted", true},
```
lack, line 256: Other test case ideas:
- What if tmpdir does not exist (Or is that relevant at this level?)

pkg/compare/container_test.go
```diff
@@ -0,0 +1,326 @@
...
+	execCommand = fakeExecCommand
+	defer func() { execCommand = exec.Command }()
+
+	tests := []struct {
+		path                string
+		tempContainerRefDir string
+		expectError         bool
+	}{
+		{"container://image:tag:/etc/configs", "/tmp/refdir", false},
+		{"invalid-path", "/tmp/refdir", true},
```
lack, line 308: We should think about adding a code coverage measurement to our unit testing - But think about how many of the interesting if clauses in the function-under-test are being exercised by the test cases...

README.md
```diff
@@ -56,6 +56,12 @@ Run a known valid reference configuration with an `oc must-gather` output:
 kubectl cluster-compare -r ./reference/metadata.yaml -f "must-gather*/*/cluster-scoped-resources","must-gather*/*/namespaces" -R
 ```
 
+Extract a known valid reference configuration from a container image and compare with a local set of CRs:
```
lack, line 59:
```suggestion
Extract a reference configuration from a container image and compare with a local set of CRs:
```

pkg/compare/compare.go
```diff
@@ -88,6 +88,9 @@ var (
 
 		# Run a known valid reference configuration with a must-gather output:
 		kubectl cluster-compare -r ./reference/metadata.yaml -f "must-gather*/*/cluster-scoped-resources","must-gather*/*/namespaces" -R
+
+		# Extract a known valid reference configuration from a container image and compare with a local set of CRs:
```
lack, line 92:
```suggestion
		# Extract a reference configuration from a container image and compare with a local set of CRs:
```

pkg/compare/container.go
```diff
@@ -0,0 +1,140 @@
...
+	out, err := engine.runEngineCommand("cp", engine.containerID+":"+pathToMetadata, dname)
+	if err != nil {
+		return fmt.Errorf("could not copy templates from container: %s", out)
+	}
+	engine.tempDir = filepath.Join(dname, filepath.Base(pathToMetadata))
+	return nil
+}
+
+// cleanup stops and removes the container used to extract the reference configs.
+func (engine *engine) cleanup() error {
```
lack, line 101: This function can only return `nil` so maybe just
```suggestion
func (engine *engine) cleanup() {
```

lack (changes requested):
This is awesome!

Some minor cleanup suggestions in comments.

lack (approved):

PR #174, opened 2025-02-13
OCPBUGS-50707: Hide all reference correlation overlaps behind a single warning

lack:
The specific issues are revealed when `--verbose` output is enabled.

### DIFF
--- a/docs/user-guide.md
+++ b/docs/user-guide.md
@@ -216,15 +216,23 @@ In such scenarios take these steps:
 
 #### There can be more than one Reference Design CR of the same Kind
 
-In this case you will have a warning presented before the diff output, formatted similar to this:
-`More then one template with same apiVersion, metadata_namespace, kind. By Default for each Cluster CR that is correlated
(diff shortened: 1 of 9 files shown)


docs/user-guide.md
```diff
@@ -216,15 +216,23 @@ In such scenarios take these steps:
...
-For each cluster CR the template that contains the least diffs will be used, to choose a different template for the CR
-pass a user config (-c) and specify in the user config file the template that should be matched to the CR. For info about
-the exact syntax view the user config section.
+In this case you will have a warning presented before the diff output,
+formatted similar to this: `The reference contains overlapping object
+definitions which may result in unexpected correlation results. Re-run with
+'--verbose' output enabled to view a detailed description of the issues`
+
+This means the template contains two CRs with the same
+apiversion-kind-name-namespace but different spec. In such cases For each
```
adhil0, line 225: NIT:
```suggestion
apiversion-kind-name-namespace but different spec. In such cases for each
```

PR #188, opened 2025-03-20
OCPBUGS-53409: diff tie breaker 4.18

lack:
- **Tighten up verbose messages for comparison and correlation debugging**
- **Use a diffscore and not mergepatch leafCount to decide between multiple matches**


nocturnalastro (approved):
/lgtm

PR #194, opened 2025-04-07
OCPBUGS-44249: Introduce 'doNotMatch' template function

lack:
This allows templates to actively opt-out of matching certain CRs.

For example, this can be useful for nameless templates which would
otherwise match too many CRs. if there is some other criteria for
inclusion/exclusion, the template may check and call the `doNotMatch`
function if it should not be allowed to match the current CR.

See `d1-anyname.yaml` for a contrived but useful example.


nocturnalastro (approved):
/lgtm

PR #221, opened 2025-07-17
Minimal example reference for users 

rohennes:
Opening a PR to see if a minimal reference is possible to help users try out the cluster-compare plugin quickly on any cluster.

### DIFF
--- a/docs/example/README.md
+++ b/docs/example/README.md
@@ -1,0 +1,15 @@
+# Baseline reference configuration
+
+This repo provides a simple example reference configuration for the `cluster-compare` plugin.
+
+## Goal
+
+Create a minimal baseline reference configuration that works with `oc cluster-compare` on any OpenShift cluster. This helps users get started with the cluster-compare plugin quickly and easily.
+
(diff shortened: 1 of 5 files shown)


docs/example/README.md
```diff
@@ -0,0 +1,16 @@
...
+Create a minimal baseline reference configuration that works with `oc cluster-compare` on any OpenShift cluster. This helps users get started with the cluster-compare plugin quickly and easily.
+
+## Usage
+
+Run the reference configuration against your cluster:
+
+````
+oc cluster-compare -r https://raw.githubusercontent.com/openshift/kube-compare/main/docs/example/metadata.yaml
+````
+
```
lack, line 16: I think it's complaining about this blank line at the end, too.

lack (commented):
/approve

docs/example/README.md
rohennes: Updated

This is a very minimal example too. I'm sure it could be built on with more example errors etc.

Issue #262, opened 2026-03-04
Add an option to generate a reference configuration from a running cluster


dcritch:
It would be useful to have the ability to capture resources from a running Kubernetes cluster and generate a reference directory + metadata.yaml compatible with the tool. It would allow for easier comparisons between different running clusters, as well as generating manifests from a cluster in a known, good state. A configuration file can be supplied so that only desired objects are captured in the output artifacts.

dcritch:
Fixed by #267

PR #267, opened 2026-03-17
Add an option to generate a reference configuration from a running cluster

dcritch:
This PR implements a method to generate a reference config from a running cluster or must-gather output. It addresses #262

Assisted by Claude Code/Cursor

## Summary by CodeRabbit

* **New Features**
  * Added a reference-generation mode to capture resources from live clusters or must-gather outputs, with CLI flags for config/output dir, per-resource selection, sanitization, omission rules, and generation summaries.

* **Documentation**
  * Added example generate-config manifest and a detailed guide on generating references and resulting output structure.

* **Tests**
  * Added unit and end-to-end tests for config loading, fetchers (cluster/must-gather), sanitization, and generator output.

* **Chores**
  * Promoted YAML library to a direct dependency.


pkg/compare/compare.go
```diff
@@ -74,6 +75,11 @@ var (
 
 		Note: KUBECTL_EXTERNAL_DIFF, if used, is expected to follow that convention.
 
+		Generate mode: Use -g with a generate config file to create a reference from a live cluster
+		or must-gather directory. The config specifies which resource types to capture. Use
+		--must-gather to generate from a must-gather directory instead of a live cluster.
```
imiller0: Can we use -f in the same way it is used for doing comparisons. This would eliminate having a separate set of options for each usage mode.

pkg/generate/generator.go
```diff
@@ -0,0 +1,218 @@
...
+	}
+	absPath, err := filepath.Abs(g.outputDir)
+	if err != nil {
+		return g.outputDir, nil
+	}
+	return absPath, nil
+}
+
+func (g *Generator) writeCRFiles(spec *ResourceSpec, resources []*unstructured.Unstructured) error {
+	kindDir := filepath.Join(g.outputDir, spec.Kind)
```
imiller0: This passes the Kind and outputDir directly from user input into a path that will be opened. The filenames should be sanitized to eliminate possibility of going outside current directory

imiller0 (commented):
Thanks for the PR and extended functionality. A lot of good stuff here. I've left a few comments in the code but also a couple high level thoughts.
1. Can you add go test code to cover the new functionality
2. As much as possible we should use common CLI flags for functionality to make the feature appear more integrated.
3. There's a couple places where additional user configurability will (I think) be required to maintain a generate / re-generate workflow where the reference is periodically updated and they won't want to have to make the same updates/edits to the captured reference (eg default fields to omit).
Thanks!

dcritch:
> Thanks for the PR and extended functionality. A lot of good stuff here. I've left a few comments in the code but also a couple high level thoughts.
> 
> 1. Can you add go test code to cover the new functionality
> 2. As much as possible we should use common CLI flags for functionality to make the feature appear more integrated.
> 3. There's a couple places where additional user configurability will (I think) be required to maintain a generate / re-generate workflow where the reference is periodically updated and they won't want to have to make the same updates/edits to the captured reference (eg default fields to omit).
>    Thanks!

Thanks Ian! I'll pushed some changes to address 2, and will continue to work on your other points over the next few days.

dcritch:
hey @imiller0 I think I covered most of your comments. please take a look and let me know what you think

dcritch:
hi @imiller0 is there anything else that needs to be done to get this PR merged?

pkg/generate/generator.go
```diff
@@ -0,0 +1,457 @@
...
+	}
+	return safe
+}
+
+// cleanResource returns a deep copy of the object (via DeepCopy) with runtime-managed fields removed.
+// fto is the merged fieldsToOmit map (defaults entries drive annotation/label key removal).
+func cleanResource(obj *unstructured.Unstructured, fto map[string]any) map[string]any {
+	result := obj.DeepCopy().Object
+	if metadata, ok := result["metadata"].(map[string]any); ok {
+		for _, key := range []string{"resourceVersion", "uid", "creationTimestamp", "generation", "managedFields", "selfLink"} {
```
lack: Is hard-coding this list safe?  Is there a way for us to query the server to learn what the full list of runtime-managed fields is, in case more are added in some future version of k8s?

dcritch: hard-coding seems to be generally safe, however I did notice some duplication of effort between the compare and the generate functions, so I've consolidated them in to one spot that can be shared amongst the code base.

dcritch:
@lack @imiller0 is there anything else needed to move this forward?

lack (approved):

dcritch:
:tada: thanks @lack !

PR #269, opened 2026-04-14
fix: correct JSON field name typo "ValidationIssuses" → "ValidationIssues"

### LINKED ISSUE #268: JSON output typo: "ValidationIssuses" should be "ValidationIssues"
The kube-compare JSON output contains a typo in the Summary object. The field is named `ValidationIssuses` (with an extra "s") instead of the correct spelling `ValidationIssues`.

**Location:**
- File: `pkg/compare/output.go`
- Line: 77
- Current code:
```go
ValidationIssues       map[string]map[string]ValidationIssue `json:"ValidationIssuses"`
```

**Expected:**
```go
ValidationIssues       map[string]map[string]ValidationIssue `json:"ValidationIssues"`
```

**Impact:**
- Minor: The typo exists in the JSON tag only, not in the Go field name
- The internal code uses the correct `ValidationIssues` field name
- External tools consuming the JSON output must use the misspelled key

**Request:**
Please fix this typo if possible. We understand this may require consideration of backward compatibility with existing consumers of the JSON output.

**Verification:**
Checked in latest main branch (as of 2026-04-13) - typo still present.

Yanhu007:
Fixes #268

## Problem

The `Summary` struct has a JSON tag typo: `"ValidationIssuses"` (extra "s") instead of `"ValidationIssues"`. This affects all JSON and YAML output, making it harder for downstream consumers to parse the field correctly.

## Changes

- Fixed the JSON tag in `pkg/compare/output.go`
- Updated all test fixtures (golden files) to match
- Updated documentation in `docs/user-guide.md`

⚠️ **Breaking change**: Any existing JSON parsers using the misspelled field name will need to update to the corrected spelling.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a typo in the output reports: the summary field is now correctly named "ValidationIssues" (previously "ValidationIssuses") across JSON and YAML outputs.
* **Documentation**
  * Updated example output in documentation to use the corrected "ValidationIssues" field and added a note about updating parsers.
* **Tests**
  * Updated test fixtures/golden outputs to reflect the corrected field name.


lack:
Great catch!  We'll have to do some consideration about the output format change so we don't immediately break existing clients, though.

PR #271, opened 2026-04-15
WIP: Add pkg/rdsdiff: reusable library for RDS version diffing

Tal-or:
Add a reusable library for comparing two versions of a telco RDS (Reference Design Specification). Given two downloadable RDS sources (GitHub tree URLs or direct archive links), the library handles the full workflow: download and extract archives, run the PolicyGenerator binary to produce policies, extract individual CRs from the generated ACM Policy documents, normalize and diff the CRs, and produce both a human-readable text report and a structured JSON comparison summary.

This package is intended to be consumed by tools such as kube-compare-mcp, which can wrap it as a thin MCP tool handler without embedding domain logic.

NOTE:
At the moment the tool can perform diffs only for telco-ran RDS

Signed-off-by: Talor Itzhak <titzhak@redhat.com>
Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

## Summary by CodeRabbit

* **New Features**
  * Extract embedded policy resources from generated YAML, normalize/serialize them deterministically, correlate old vs new sets, and produce human-readable diff reports plus a structured JSON summary.
  * Download and safely extract remote archives (zip/tar.gz), detect effective roots, and validate configuration layouts.
  * Session-scoped workspace helpers and a policy-generation runner to produce generated policy artifacts.
  * Added exported constants for policy source paths and a small policy list.

* **Tests**
  * Comprehensive tests covering extraction, normalization, diffing, download/extract, session handling, policy generation, and report/JSON outputs.

### DIFF
--- a/pkg/rdsdiff/compare.go
+++ b/pkg/rdsdiff/compare.go
@@ -1,0 +1,502 @@
+// SPDX-License-Identifier:Apache-2.0
+
+package rdsdiff
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"os"
+	"path/filepath"
+	"regexp"
+	"sort"
+	"strings"
+
+	"github.com/sergi/go-diff/diffmatchpatch"
+	sigsyaml "sigs.k8s.io/yaml"
+)
+
+var slashRe = regexp.MustCompile(`/+`)
+
+// CRKeyFromObject builds a stable key for a CR: kind_namespace_name, or kind_name for cluster-scoped.
+func CRKeyFromObject(obj map[string]any) string {
+	kind, _ := obj["kind"].(string)
+	if kind == "" {
+		kind = "Unknown"
+	}
+	meta, _ := obj["metadata"].(map[string]any)
+	if meta == nil {
+		meta = make(map[string]any)
+	}
+	name, _ := meta["name"].(string)
+	if name == "" {
+		name = "unnamed"
+	}
+	ns, _ := meta["namespace"].(string)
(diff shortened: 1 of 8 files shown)


lack:
For a new feature like this, can we add a markdown document to the docs/ showing how and why this will be used, maybe with examples?

Tal-or:
> For a new feature like this, can we add a markdown document to the docs/ showing how and why this will be used, maybe with examples?

Sure thing, I'll add that, Thanks!

lack:
The concern here is that we don't want to introduce anything telco-specific (or RDS-specific, since that it telco-specific), as this is intended to be a generic tool.

If this is a generic feature, a document explaining its purpose, examples, and how it works with the existing features of this tool would be a great way to understand it.

Tal-or:
@lack this feature can be implement in a way it can be used to find diff between any RDS type.
for example the tool will accept 2 container images containing RDS ref from the same type, compare between them in a generic way and find the diffs between the two versions. 

This is similar to what the standard kube-compare is doing but instead of having 
ref vs live cluster 
it's
ref v1 vs ref v2
WDYT?

lack:
When we're comparing a reference versus a live cluster, the value this tool brings above a plain invocation of `diff` is that it can hide "unimportant" differences (where what parts are "unimportant" are defined in the reference templates).

When we're comparing one reference versus another reference, what does this tool add that could not already be achieved by running `diff`?

Tal-or:
Closing till we'll finalize the design

PR #280, opened 2026-06-24
OCPBUGS-92049: Remove getHostByName from sprig FuncMap

lack:
## Description

This PR fixes the `kc-cmp-001` security finding. 

`kube-compare` attempts to sandbox the template execution environment (for untrusted remote references) by stripping `env` and `expandenv` from the `sprig` function map. However, it accidentally leaves `getHostByName` available.

A malicious reference author could craft a template that reads a cluster Secret via `lookupCRs`, encodes it, and passes it to `getHostByName`. The tool would then perform a DNS lookup while rendering the template, covertly exfiltrating sensitive cluster data via DNS.

This change explicitly removes `getHostByName` from the funcmap to close this DNS egress channel.

## Testing
Tested with `make test`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Removed access to an unsafe template helper from the standard template function set, preventing templates from using it via the built-in helper map.
  * Kept existing helper mappings and their help-text behavior the same.

### DIFF
--- a/pkg/compare/funcmap.go
+++ b/pkg/compare/funcmap.go
@@ -41,6 +41,7 @@ func FuncMap() template.FuncMap {
 	f := sprig.TxtFuncMap()
 	delete(f, "env")
 	delete(f, "expandenv")
+	delete(f, "getHostByName")
 
 	for key := range f {
 		FuncHelp[key] = SprigImportFlag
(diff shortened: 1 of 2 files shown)


nocturnalastro (approved):
/lgtm

nocturnalastro (approved):

PR #282, opened 2026-07-14
Add Helm-compatible include template function

lack:
## What

Add the `include` template function, matching Helm's semantics. Unlike Go's built-in `{{template}}` action which writes directly to the output, `{{include}}` executes a named template and returns its result as a string. This allows the output to be piped through other template functions like `indent`, `upper`, `nindent`, etc.

## Why

The `include` function is one of the most commonly used Helm template functions. Reference authors coming from Helm expect it to be available, and it enables cleaner template composition — defining reusable snippets (e.g. standard labels, annotations) in `_helpers.tpl` files and including them with post-processing.

Without `include`, authors are limited to Go's `{{template}}` which cannot be piped:
```
{{/* This works but cannot be indented */}}
{{ template "mychart.labels" . }}

{{/* This is what Helm users expect */}}
{{ include "mychart.labels" . | indent 4 }}
```

## How

The implementation follows the same pattern Helm uses in `pkg/engine/engine.go`:

- A **placeholder** `include` is registered in `FuncMap()` so templates can parse without error
- After parsing, `InitInclude()` **late-binds** the real implementation as a closure over the parsed `*template.Template`, enabling it to call `ExecuteTemplate()`
- A **recursion guard** (max 1000 depth, matching Helm) prevents infinite loops from self-including templates

The function was not imported directly from Helm's engine package because that would pull in the entire chart rendering pipeline (chart loading, Kubernetes client, values handling, etc.) as a dependency. The actual `include` implementation is ~25 lines.

## Testing

- **Unit tests** (`funcmap_include_test.go`): basic include, data passing, piping through `upper`/`indent`, nested includes, missing template error, recursion guard
- **Integration test** (`TestCompareRun/Include_Function`): end-to-end test with a `_helpers.tpl` defining labels and a ConfigMap template using `{{ include "mychart.labels" . | indent 4 }}`

## Summary by CodeRabbit

- **New Features**
  - Added Helm-style `include` support for templates, enabling named partial reuse, passing data context, and using included output with built-in template functions.
  - Added protection against infinite self-referential includes via an enforced recursion limit.
  - Missing included templates now surface clear execution errors.
- **Tests**
  - Expanded coverage for basic, nested, formatted/piped includes, missing-template errors, and recursion-guard behavior, including updated fixtures.
- **Documentation**
  - Updated the config guide to document `include`, including usage examples and the recursion-guard behavior.

### DIFF
--- a/docs/reference-config-guide-v1.md
+++ b/docs/reference-config-guide-v1.md
@@ -84,7 +84,7 @@ inclusion/exclusion of content. The following types of user variation are expect
 
 GO templating allows use of custom and built in functions to allow complex use cases. In this version all Go built-in
 functions are supported along with the functions in the Sprig library. Also this version follows the Helm templating
-behavior and supports all custom functions that are used in helm (example: toYaml).
+behavior and supports several Helm-style helper functions (e.g. toYaml, toJson, fromYaml, fromJson, include).
 
 ```yaml
 apiVersion: v1
--- a/docs/reference-config-guide-v2.md
+++ b/docs/reference-config-guide-v2.md
@@ -148,7 +148,7 @@ inclusion/exclusion of content. The following types of user variation are expect
 
 GO templating allows use of custom and built in functions to allow complex use cases. In this version all Go built-in
 functions are supported along with the functions in the Sprig library. Also this version follows the Helm templating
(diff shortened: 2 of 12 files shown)


nocturnalastro (approved):
/lgtm

